PR #14, opened 2019-10-15
Make sample qc exprs return an array instead of a map

henrydavidge:
 Now that https://github.com/apache/spark/pull/25666 has merged, it's more convenient and performant to return an array and filter by index. In addition, we can now generate sample qc stats when we don't have sample ids, which we expect to be the common case.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/sql/expressions/PerSampleSummaryStatistics.scala
+++ b/core/src/main/scala/io/projectglow/sql/expressions/PerSampleSummaryStatistics.scala
@@ -56,8 +56,9 @@ case class PerSampleSummaryStatistics(
   override def nullable: Boolean = false
 
   override def dataType: DataType =
-    if (optionalFieldIndices(0) != -1) {
-      ArrayType(MomentAggState.schema.add(VariantSchemas.sampleIdField))
+    if (hasSampleIds) {
+      val fields = VariantSchemas.sampleIdField +: MomentAggState.schema.fields
+      ArrayType(StructType(fields))
     } else {
       ArrayType(MomentAggState.schema)
     }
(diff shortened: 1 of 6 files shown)


core/src/main/scala/io/projectglow/sql/expressions/PerSampleSummaryStatistics.scala
```diff
@@ -54,38 +54,52 @@ case class PerSampleSummaryStatistics(
 
   override def children: Seq[Expression] = Seq(genotypes)
   override def nullable: Boolean = false
-  override def dataType: DataType = MapType(StringType, MomentAggState.schema)
+
+  override def dataType: DataType =
+    if (optionalFieldIndices(0) != -1) {
+      ArrayType(MomentAggState.schema.add(VariantSchemas.sampleIdField))
```
karenfeng: OOC, why do we prepend sample ID here but append for the call summary stats?

henrydavidge: Good point, better to be consistent. I'll make it the first spot in each.

core/src/test/scala/io/projectglow/tertiary/SampleQcExprsSuite.scala
```diff
@@ -87,19 +84,19 @@ class SampleQcExprsSuite extends GlowBaseTest {
     spark.createDataFrame(data)
   }
 
-  private def readVcf(path: String): DataFrame = {
+  private def readVcf(path: String, includeSampleIds: Boolean = true): DataFrame = {
```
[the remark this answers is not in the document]
henrydavidge: Done

PR #29, opened 2019-10-19
Include the Git Hash in the MANIFEST file

ahirreddy:
Adds the Git hash to the MANIFEST file. We do so by adding it directly to the Manifest attributes in the `core` project. We shell out to git to get the current HEAD commit.


karenfeng:
I've manually verified that `Git-Release-Hash` is included in the manifest. Merging.

Issue #32, opened 2019-10-21
Source in package org.bdgenomics.adam.rdd not necessary


heuermh:
[`VCFMetadataLoader`](https://github.com/projectglow/glow/blob/master/core/src/main/scala/org/bdgenomics/adam/rdd/VCFMetadataLoader.scala) does not appear to use `ADAMContext`, so I'm not sure why it needs to be in the `org.bdgenomics.adam.rdd` package

```scala
package org.bdgenomics.adam.rdd

import htsjdk.variant.vcf.VCFHeader
import org.apache.hadoop.conf.Configuration
import org.apache.hadoop.fs.Path
import org.seqdoop.hadoop_bam.util.{VCFHeaderReader, WrapSeekable}

object VCFMetadataLoader {

  // Currently private in ADAM (ADAMContext).
  def readVcfHeader(config: Configuration, pathName: String): VCFHeader = {
    val is = WrapSeekable.openPath(config, new Path(pathName))
    val header = VCFHeaderReader.readHeaderFrom(is)
    is.close()
    header
  }
}
```

karenfeng:
Thanks for the catch @heuermh! I addressed this in https://github.com/projectglow/glow/pull/35.

PR #35, opened 2019-10-21
Move VCFMetadataLoader inside glow

karenfeng:

Addresses https://github.com/projectglow/glow/issues/32: we no longer use any package-private code from `org.bdgenomics.adam.rdd`, so it's safe to move `VCFMetadataLoader` to inside `io.projectglow`.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


core/src/main/scala/io/projectglow/vcf/VCFMetadataLoader.scala
```diff
@@ -23,7 +23,7 @@ import org.seqdoop.hadoop_bam.util.{VCFHeaderReader, WrapSeekable}
 
 object VCFMetadataLoader {
 
-  // Currently private in ADAM (ADAMContext).
+  // From ADAMContext
```
henrydavidge: nit: link to the source file on github?

PR #44, opened 2019-10-23
Fix notebooks in the docs

kianfar77:
Fixes notebooks in the docs (switch to glow, fixes paths, removed dependencies on databricks) 
Explanations on notebooks.
Minor edits. 

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ x] Manual tests

(Details)


docs/source/getting-started.rst
```diff
@@ -49,10 +49,18 @@ The easiest way to use Glow in the cloud is with the `Databricks Runtime for Gen
 provider or Spark distribution. You need to install the maven package
 ``io.project:glow_2.11:${version}`` and optionally the Python frontend ``glow.py``.
 
-Example notebook
-----------------
+Notebooks embedded in the docs
```
karenfeng: I added a similar section for this here: https://github.com/projectglow/glow/blob/master/docs/source/additional-resources.rst (there is also a hard-coded link to this in the notebook extension: https://github.com/projectglow/glow/blob/master/docs/extensions/notebook.py). We should probably choose which of these to point to. Did we remove any reference to `dbutils` as well?

docs/source/getting-started.rst
```diff
@@ -49,10 +49,18 @@ The easiest way to use Glow in the cloud is with the `Databricks Runtime for Gen
...
+------------------------------
+
+To demonstrate example use cases of Glow functionalities, most of the doc pages on this website are accompanied by embedded notebooks.
+The commands in these notebooks are cloud-independent, except for the ``display()`` command, which is specific to `Databricks Notebooks <https://docs.databricks.com/notebooks/index.html>`_ but can be easily replaced by other commands like ``.show``.
```
karenfeng: `df.show` does not work in Python - you'll need the close-paren.

docs/source/etl/lift-over.rst
```diff
@@ -66,7 +66,7 @@ Pull a target assembly reference file down to every node in the Spark cluster in
 performing variant liftover.
 
 The ``lift_over_variants`` transformer operates on a DataFrame containing genetic variants and supports the following
-:ref:`options <transformer-options>`.
+options:
```
karenfeng: I think I resolved this in master - https://github.com/projectglow/glow/blob/master/docs/source/tertiary/pipe-transformer.rst

kianfar77:
> In the notebooks, please replace `"com.databricks.vcf"` with `"vcf"`. Also make sure to test the notebooks on DBR with Glow attached rather than the Genomics Runtime - there are additional steps required such as `import glow` (on that note, please remove any references to `db_genomics`). The notion of using environment variables to install a reference genome on a cluster should also be dropped, or at least explained in the docs. We should also edit `regression-tests.html`, which contains the Databricks logo, references to "HLS Runtime," and so on.

Already done as we spoke. Fixed a few missed items.

docs/source/getting-started.rst
kianfar77: removed

docs/source/getting-started.rst
kianfar77: edited and made a reference to your paragraph

docs/source/etl/lift-over.rst
kianfar77: Not sure why this "options" should refer to pipe transformer options.

karenfeng: Perhaps it'd make sense to have a separate section describing transformers, as we expose the same Scala/Python APIs and share behavior such as case-sensitivity.

kianfar77: Sure will do in a separate PR

PR #49, opened 2019-10-25
Function to aggregate across all elements at each index in array columns

henrydavidge:

We currently include a few functions to compute sample-based quality control statistics. However, users can't write their own functions that aggregate across the same dimension without getting intimately familiar with Spark internals.

This PR adds a function,  `aggregate_by_index`, that provides the scaffolding for users to write their own aggregations. The user provides an initial value, an update function to merge a buffer object with a new datum, a merge function to combine two buffers, and optionally an evaluate function called on the buffer to produce a result. These arguments are analogous to those of the `aggregate` higher order function built into Spark: https://docs.databricks.com/_static/notebooks/apache-spark-2.4-functions.html.

I'll provide documentation for the new function as well as the initialization changes in a separate PR. 

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/sql/SqlExtensionProvider.scala
+++ b/core/src/main/scala/io/projectglow/sql/SqlExtensionProvider.scala
@@ -19,28 +19,31 @@ package io.projectglow.sql
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.FunctionIdentifier
 import org.apache.spark.sql.catalyst.analysis.{FunctionRegistry, UnresolvedExtractValue}
-import org.apache.spark.sql.catalyst.expressions.CreateNamedStruct
+import org.apache.spark.sql.catalyst.expressions.{CreateNamedStruct, LambdaFunction}
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.internal.SQLConf
-
 import io.projectglow.common.VariantSchemas
 import io.projectglow.sql.expressions._
-import io.projectglow.sql.optimizer.HLSReplaceExpressionsRule
+import io.projectglow.sql.optimizer.{HLSReplaceExpressionsRule, ResolveAggregateFunctionsRule}
+
+// TODO(hhd): Spark 3.0 allows extensions to register functions. After Spark 3.0 is released,
+// we should move all extensions into this class.
+class GlowSQLExtensions extends (SparkSessionExtensions => Unit) {
+  val resolutionRules: Seq[Rule[LogicalPlan]] = Seq(ResolveAggregateFunctionsRule)
+  val optimizations: Seq[Rule[LogicalPlan]] = Seq(HLSReplaceExpressionsRule)
+  def apply(extensions: SparkSessionExtensions): Unit = {
+
+    resolutionRules.foreach(r => extensions.injectResolutionRule(_ => r))
+    optimizations.foreach(r => extensions.injectOptimizerRule(_ => r))
+  }
+}
 
 object SqlExtensionProvider {
 
   final def register(session: SparkSession): Unit = {
-    val experimental = session.experimental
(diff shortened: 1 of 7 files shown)


core/src/test/scala/io/projectglow/sql/GlowBaseTest.scala
```diff
@@ -45,15 +45,18 @@ abstract class GlowBaseTest
...
-    SparkSession.setActiveSession(session)
+  override def initializeSession(): Unit = ()
+
+  override protected implicit def spark: SparkSession = {
+    val sess = SparkSession.builder().config(sparkConf).master("local[2]").getOrCreate()
+    SqlExtensionProvider.register(sess)
+    SparkSession.setActiveSession(sess)
     Log.setGlobalLogLevel(Log.LogLevel.ERROR)
-    session
+    sess
```
henrydavidge: I had to change our test harness to create a `SparkSession` this way because the default `TestSparkSession` doesn't load extensions :(.

core/src/main/scala/io/projectglow/sql/SqlExtensionProvider.scala
```diff
@@ -146,5 +149,16 @@ object SqlExtensionProvider {
       FunctionIdentifier("vector_to_array"),
       exprs => VectorToArray(exprs.head)
     )
+
+    functionRegistry.registerFunction(
+      FunctionIdentifier("aggregate_by_index"),
```
karenfeng: Tiny nit: can we restore the function ordering in this file? I think it's easiest to visually parse whether a function exists that way.

core/src/main/scala/io/projectglow/sql/optimizer/hlsOptimizerRules.scala
```diff
@@ -16,13 +16,14 @@
 
 package io.projectglow.sql.optimizer
 
+import io.projectglow.sql.expressions.{AddStructFields, AggregateByIndex, UnwrappedAggregateFunction}
```
karenfeng: Tiny nit: can we restore the original ordering?

core/src/main/scala/org/apache/spark/sql/SQLUtils.scala
```diff
@@ -20,6 +20,9 @@ import org.apache.spark.TaskContext
 import org.apache.spark.ml.linalg.{MatrixUDT, VectorUDT}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.analysis.SimpleAnalyzer
```
karenfeng: Nit: do we need this imports?

core/src/test/scala/io/projectglow/sql/GlowBaseTest.scala
```diff
@@ -43,17 +43,20 @@ abstract class GlowBaseTest
       .set("spark.kryo.registrationRequired", "false")
       .set(
         "spark.hadoop.io.compression.codecs",
-        "org.seqdoop.hadoop_bam.util.BGZFCodec,org.seqdoop.hadoop_bam.util.BGZFEnhancedGzipCodec"
+        classOf[BGZFCodec].getCanonicalName
```
karenfeng: Do we not need `BGZFEnhancedGzipCodec` here? I think we definitely have `.gz` files.

henrydavidge: It was actually an error to include it there. We explicitly add the "enhanced" codec when reading VCF files. We also do not enable the enhanced codecs for all datasources in the genomics runtime: https://docs.databricks.com/release-notes/runtime/3.2.html#changes-and-improvements.

core/src/main/scala/io/projectglow/sql/SqlExtensionProvider.scala
karenfeng: Bump - I think it makes sense to keep `SqlExtensionProvider` sorted.

PR #59, opened 2019-11-05
Add BGEN option to explicitly not include sample ids

henrydavidge:

Previously, if a BGEN file contained sampleIds, they were always included in the emitted DataFrame. If a user didn't want them (common when dealing with large cohort sizes), they had to manually drop them.

Now, you can provide the same `includeSampleIds` option used by the VCF and Plink datasources. If it is `false`, sample ids are not included in the output.

cc @emaxwell 

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/bgen/BgenFileFormat.scala
+++ b/core/src/main/scala/io/projectglow/bgen/BgenFileFormat.scala
@@ -36,9 +36,8 @@ import org.skife.jdbi.v2.DBI
 import org.skife.jdbi.v2.util.LongMapper
 
 import io.projectglow.common.logging.{HlsMetricDefinitions, HlsTagDefinitions, HlsTagValues, HlsUsageLogging}
-import io.projectglow.common.{GlowLogging, WithUtils}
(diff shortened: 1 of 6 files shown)


core/src/main/scala/io/projectglow/bgen/BgenFileFormat.scala
```diff
@@ -38,6 +38,7 @@ import org.skife.jdbi.v2.util.LongMapper
 import io.projectglow.common.logging.{HlsMetricDefinitions, HlsTagDefinitions, HlsTagValues, HlsUsageLogging}
 import io.projectglow.common.{GlowLogging, WithUtils}
 import io.projectglow.sql.util.{ComDatabricksDataSource, SerializableConfiguration}
+import io.projectglow.vcf.VCFOption
```
kianfar77: It may be a good idea to gather all bgen options (including this one) in a BGENOption object like what we did for VCFOption. It will be cleaner and creates separation of VCF and BGEN. Just a thought. Let me know if you disagree and I will approve.

PR #64, opened 2019-11-07
Fix mvn version

karenfeng:

Adds a release step that sets the latest stable version in a new version file (`stable-version.sbt`). This is used to keep the docs up to date.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual tests

Running `sbt release` creates or updates `stable-version.sbt`, and the docs contain the version in `stable-version.sbt` instead of `version.sbt`.


henrydavidge:
I don't think this is right, actually. The version in `version.sbt` should reflect the version that we would push if we deployed right now, which would be a preview (`-SNAPSHOT`) of 1.0.3.

henrydavidge:
For example, see `version.sbt` in Delta: https://github.com/delta-io/delta/blob/master/version.sbt

stable-version.sbt
```diff
@@ -0,0 +1 @@
+version in ThisBuild := "0.1.2"
```
henrydavidge: nit: if we're not reading this from sbt, might be easier to just make this a text file (like the content is only `0.1.2`).

PR #65, opened 2019-11-07
More changes for logging and bgen write options

kianfar77:
Adds minor changes to work with logging 
Put bigbegn write options in common.BgenOptions


core/src/main/scala/io/projectglow/transformers/pipe/PipeTransformer.scala
```diff
@@ -141,6 +141,7 @@ object PipeTransformer {
   private val ENV_PREFIX = "env_"
   private val INPUT_FORMATTER_PREFIX = "in_"
   private val OUTPUT_FORMATTER_PREFIX = "out_"
+  val LOGGING_BLOB_KEY = "pipeCmdTool"
```
karenfeng: Why are we moving this here?

kianfar77: For all other features, like vcf/bgen/plink read/write and variant normalizer transformer, the options which are defined inside the respective packages are used as keys for the maps in the logging json blobs. I do not want this one to be an exception and be defined outside of pipe transformer. This makes the logging code cleaner on the universe side as well.

karenfeng: I see. Can we move the transformer options to a higher-level object as we did for the datasource options for transparency?

kianfar77: Sure I can. But if ok with you let's do it in another PR because I want to get the universe PR review moving froward with data and observability teams and that depends on this PR. Moving options to higher level can be done independently.

karenfeng: Ok, that works.

PR #68, opened 2019-11-08
Explicitly support only Scala 2.11 and Spark 2.4.3

karenfeng:

Spark 2.4.2 (and 3.0+) is built on Scala 2.12; the other 2.4 versions (2.4.0, 2.4.1, 2.4.3) are built on Scala 2.11. To be compatible with different versions of Spark, we should eventually cross-build across 2.11 and 2.12; for now, we can explicitly support only Spark 2.4.3.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual tests

Tested running basic Glow commands (`glow.register(spark)` and `glow.read.format('vcf').load('test.vcf')`) with [PySpark 2.4.3, Glow 2.11] and [PySpark 2.4.2, Glow 2.12].


build.sbt
```diff
@@ -94,8 +98,8 @@ lazy val dependencies = Seq(
     .exclude("org.apache.commons", "commons-collections4")
     .exclude("org.apache.commons", "commons-vfs2")
     .exclude("org.apache.hadoop", "hadoop-client")
-    .exclude("org.apache.spark", s"spark-mllib_$scalaMajorMinor")
-    .exclude("org.bdgenomics.adam", s"adam-core-spark2_$scalaMajorMinor")
+    .exclude("org.apache.spark", s"spark-mllib_$scala211MajorMinor")
```
henrydavidge: Shouldn't we be excluding the 2.12 version as well?

build.sbt
```diff
@@ -227,7 +232,52 @@ def writeStableVersion(selectVersion: Versions => String): ReleaseStep = { st: S
   st
 }
 
+// Commit stable version
```
henrydavidge: Why this change?

karenfeng: This simplifies the release so that `stable-version.txt` is not uncommitted before we push the new version. I don't really mind if this sticks around here, but it fits into the `sbt-release` pattern of modifying files with a version and immediately committing.

docs/source/getting-started.rst
```diff
@@ -4,12 +4,11 @@ Getting Started
 Running Locally
 ---------------
 
-Glow requires Apache Spark 2.4.2 or above. If you don't have a local Apache Spark installation,
-you can install it from PyPI:
+Glow requires Apache Spark 2.4.3. If you don't have a local Apache Spark installation, you can install it from PyPI:
```
henrydavidge: (or a later version of Spark 2.4.x that is built on scala 2.11)

build.sbt
henrydavidge: Can we 1) make this a function to commit a set of files 2) move it to separate file? I don't really like having functions inline with the build definition.

henrydavidge:
Btw, AFAIK, Spark 3.0 supports _only_ Scala 2.12. So, when Spark 3 is released, we don't need to start crossbuilding. We just need to build for 2.12.

build.sbt
```diff
@@ -203,40 +202,22 @@ ThisBuild / pomIncludeRepository := { _ =>
   false
 }
 ThisBuild / publishMavenStyle := true
+ThisBuild / releaseCrossBuild := true
```
henrydavidge: Should we remove this now?

PR #79, opened 2019-11-19
Improve VCF piper performance

karenfeng:

- VCF output formatter uses a [buffered line reader](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/tribble/readers/SynchronousLineReader.java) instead of the [HTSJDK ASCII line reader](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/tribble/readers/AsciiLineReader.java) as we don't need a position-aware reader, and the latter is slower.
- Persists intermediary RDD on disk instead of in memory to avoid exorbitant GC.
- Projects from a GenericInternalRow to an [UnsafeRow](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/expressions/UnsafeRow.java) before persisting to disk for faster Kryo serialization.
- Supports stringent validation in the VCF output formatter.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual tests

For the following tests, I used the 1KG chr22 VCF ingested to Delta as the input DataFrame for the piper.

Piped script:
```
"#!/bin/sh
set -e

export tmpfile=$(mktemp -t vcf.XXXXXX)
cat - > $tmpfile
sleep 240

cat $tmpfile | grep -v "#INFO"
rm $tmpfile
```

The following code was used for schema inference:
```
import io.projectglow.Glow
val outputDf = Glow.transform("pipe", inputDf, Map(
  "cmd" -> cmd,
  "inputFormatter" -> "vcf",
  "outputFormatter" -> "vcf",
  "inValidationStringency" -> "STRICT",
  "outValidationStringency" -> "STRICT",
  "inVcfHeader" -> "infer"))
```

The following code was used for materialization:
```
outputDf.write.format("noop").save("noop")
```

Runtime for schema inference drops from 14.12m to 13.41m by using the buffered line reader.

As we have already implemented, we should cache the intermediary RDD because otherwise we have to re-perform the task from scratch; this gets expensive if the user wants to perform multiple operations and also occurs between getting the schema and actually materializing the DataFrame. We should cache the serialized data at the disk level as, generally speaking, no useful VCFs are small enough to fit cached (especially deserialized) in memory. Attempting to cache in memory, as we currently do, results in exorbitant amounts of GC.

  | None | Memory | Memory and Disk | Disk
-- | -- | -- | -- | --
Get schema | 5.02m | 13.41m | 20.11m | 10.44m
Materialize | 5.66m | 7.01m | 1.94m | 1.98m
Display | 5.98m | 1.92s | 28.87s | 3.05s

Profiling demonstrates that the runtime of the piper caching on disk and materializing the DataFrame has the following breakdown, for which the vast majority of the time is spent on serialization to disk.

- Pipe iterator write input: 19.50%
- VCF input formatter write: 11.44%
- 50/50 between InternalRow → VariantContext conversion and HTSJDK write
- Codegen’d iterator: 8.06%
- Pipe iterator read input: 8.62%
- Kryo serialization: 60.23%
- Write: 9.88%

Using the Kryo serializable class `UnsafeRow` with the disk-persisting piper thus speeds this up even further, resulting the following (final) timing results:

- Get schema: 5.88m
- Materialize: 6.01s
- Display: 1.70s


core/src/main/scala/io/projectglow/transformers/pipe/Piper.scala
```diff
@@ -70,7 +72,7 @@ private[projectglow] object Piper extends GlowLogging {
       .mapPartitions { it =>
         new PipeIterator(cmd, env, it, informatter, outputformatter)
       }
-      .cache()
+      .persist(StorageLevel.DISK_ONLY)
```
henrydavidge: You might want to try `MEMORY_AND_DISK`. I believe it works like `DISK_ONLY` but keeps blocks in memory if they fit.

core/src/main/scala/io/projectglow/vcf/VCFInputFormatter.scala
```diff
@@ -48,11 +48,7 @@ class VCFInputFormatter(
   }
 
   override def write(record: InternalRow): Unit = {
-    val vcOpt = converter.convert(record)
-    if (vcOpt.isDefined) {
-      val vc = vcOpt.get
-      writer.write(vc)
-    }
+    converter.convert(record).foreach(writer.write)
```
henrydavidge: No performance penalty for the lambda function here?

core/src/main/scala/io/projectglow/vcf/VCFOutputFormatter.scala
```diff
@@ -19,34 +19,38 @@ package io.projectglow.vcf
...
     val internalRowIter: Iterator[InternalRow] = new Iterator[InternalRow] {
-      private val projection = UnsafeProjection.create(schema)
-      private var nextRecord: InternalRow = null
+      private var nextRecord: InternalRow = _
       private def readNextVc(): Unit = {
         while (nextRecord == null && lineIterator.hasNext) {
           val decoded = codec.decode(lineIterator.next())
           if (decoded != null) {
-            nextRecord = converter.convertRow(decoded, isSplit = false).copy()
+            nextRecord = projection.apply(converter.convertRow(decoded, isSplit = false))
```
henrydavidge: nit: you can just do `project(convert...)`

core/src/main/scala/io/projectglow/vcf/VCFInputFormatter.scala
karenfeng: I don't think it's an issue for `Option`s; the main perf issues are for Lists/Maps/etc.

PR #81, opened 2019-11-20
Fix for bigvcf writer with a header but no rows

kianfar77:

#67 results in an empty vcf file if an empty dataframe is passed along with a determined header to bigvcf writer. This PR adds a condition to fix this, i.e., make sure the header is written although the DataFrame is empty. Also if number of partitions is 0 (which can happen) an appropriate message is emitted. 

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

### DIFF
--- a/core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
+++ b/core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
@@ -47,6 +47,11 @@ object BigVCFDatasource extends HlsUsageLogging {
     val rdd = data.queryExecution.toRdd
 
     val nParts = rdd.getNumPartitions
+
+    if (nParts == 0) {
+      throw new SparkException("The DataFrame is empty and has zero partitions. Cannot write vcf.")
+    }
+
     val conf = VCFFileFormat.hadoopConfWithBGZ(data.sparkSession.sparkContext.hadoopConfiguration)
     val serializableConf = new SerializableConfiguration(conf)
     val firstNonemptyPartition =
@@ -58,6 +63,7 @@ object BigVCFDatasource extends HlsUsageLogging {
         .contains(VCFFileWriter.INFER_HEADER))) {
       throw new SparkException("Cannot infer header for empty VCF.")
     }
+
     rdd.mapPartitionsWithIndex {
       case (idx, it) =>
         val conf = serializableConf.value
--- a/core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
+++ b/core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
@@ -48,25 +48,23 @@ object BigVCFDatasource extends HlsUsageLogging {
 
     val nParts = rdd.getNumPartitions
 
+    if (nParts == 0) {
+      throw new SparkException(
+        "Cannot write vcf because the DataFrame has zero partitions. " +
+        "Repartition to a positive number of partitions if you want to just write the header")
+    }
+
     val conf = VCFFileFormat.hadoopConfWithBGZ(data.sparkSession.sparkContext.hadoopConfiguration)
     val serializableConf = new SerializableConfiguration(conf)
     val firstNonemptyPartition =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -47,17 +47,26 @@ object BigVCFDatasource extends HlsUsageLogging {
...
-        .contains(VCFFileWriter.INFER_HEADER))) {
-      throw new SparkException("Cannot infer header for empty VCF.")
+    if (firstNonemptyPartition == -1) {
+      if (!options.contains(VCFFileWriter.VCF_HEADER_KEY) ||
+        options
+          .get(VCFFileWriter.VCF_HEADER_KEY)
+          .contains(VCFFileWriter.INFER_HEADER)) {
+        throw new SparkException("Cannot infer header for empty VCF.")
+      }
+      if (nParts == 0) {
```
karenfeng: I think we can throw this earlier (before we check for non-empty partitions).

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -71,8 +80,10 @@ object BigVCFDatasource extends HlsUsageLogging {
         DatabricksBGZFOutputStream.setWriteEmptyBlockOnClose(outputStream, idx == nParts - 1)
 
         // Write the header if this is the first nonempty partition
+        val partitionWithHeader = if (firstNonemptyPartition == -1) 0 else firstNonemptyPartition
```
karenfeng: Nice catch!

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -47,17 +47,26 @@ object BigVCFDatasource extends HlsUsageLogging {
...
-      throw new SparkException("Cannot infer header for empty VCF.")
+    if (firstNonemptyPartition == -1) {
+      if (!options.contains(VCFFileWriter.VCF_HEADER_KEY) ||
+        options
+          .get(VCFFileWriter.VCF_HEADER_KEY)
+          .contains(VCFFileWriter.INFER_HEADER)) {
+        throw new SparkException("Cannot infer header for empty VCF.")
+      }
+      if (nParts == 0) {
+        throw new SparkException(
```
karenfeng: Do we have a test for this exception?

kianfar77: Added

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
kianfar77: Moved up

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
kianfar77: Thanks

core/src/test/scala/io/projectglow/vcf/VCFFileWriterSuite.scala
```diff
@@ -464,6 +467,53 @@ class SingleFileVCFWriterSuite extends VCFFileWriterSuite("bigvcf") {
...
+    val rewrittenDs = spark
+      .read
+      .format(readSourceName)
+      .option("vcfRowSchema", true)
+      .load(tempFile)
+
+    val truthHeader = VCFMetadataLoader.readVcfHeader(sparkContext.hadoopConfiguration, NA12878)
+    val writtenHeader = VCFMetadataLoader.readVcfHeader(sparkContext.hadoopConfiguration, tempFile)
+
+    assert(truthHeader.getMetaDataInInputOrder.equals(writtenHeader.getMetaDataInInputOrder))
```
karenfeng: One more check: can you ensure the sample match? (`truthHeader.getGenotypeSamples == writtenHeader.getGenotypeSamples`)

kianfar77: Added

PR #82, opened 2019-11-20
Persist RDDs and Datasets on disk by default

karenfeng:

Writing with the big VCF datasource results in OOMs and/or an exorbitant amount of GC when we cache in memory if the DataFrame to be written is large. Even if we cache as `MEMORY_AND_DISK`, we can still run out of heap space as executors drop off before they can fall back to caching on disk. Caching on disk is probably the safest option.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual tests

Tested with a variety of storage levels; `DISK_ONLY` is the only one that works even when there is a very limited amount of memory (a single c5.xlarge worker).

### DIFF
--- a/core/src/main/scala/io/projectglow/common/WithUtils.scala
+++ b/core/src/main/scala/io/projectglow/common/WithUtils.scala
@@ -21,6 +21,7 @@ import java.util.concurrent.locks.Lock
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.Dataset
+import org.apache.spark.storage.StorageLevel
 
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/io/projectglow/common/WithUtils.scala
```diff
@@ -71,7 +72,7 @@ object WithUtils {
   }
 
   def withCachedRDD[T, U](rdd: RDD[T])(f: RDD[T] => U): U = {
-    rdd.cache()
+    rdd.persist(StorageLevel.DISK_ONLY)
```
henrydavidge: nit: add a comment please!

PR #84, opened 2019-11-21
Smarter sample inference for VCF writing

karenfeng:

For the sharded VCF writer, throw an error if the header (including sample IDs) is inferred from the first variant context encountered but following variant contexts have new sample IDs.
The big VCF writer does not work in the case that there are mixed missing/non-missing sample IDs. If sample IDs are non-missing, determines the unique set of all sample IDs across the DataFrame and use that as the header for all partitions. If sample IDs are missing, gets the number of missing sample IDs (must match across all rows) and uses that as the header for all partitions.

In the current behavior, new sample IDs are dropped for the sharded VCF writer. The primary issue is with the big VCF writer, which can write malformed files if the inferred headers have different numbers of sample IDs; this results in suddenly dropped columns.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

To observe how this impacts performance, I rewrote a 1KG chr22 Delta table to BigVCF/VCF with the VCF header inferred. The runtime increases for the big VCF writer as we have to collect the distinct sample IDs; it also increases for the sharded VCF writer as we perform safety checks on the sample IDs to be written.

| | Single VCF | Sharded VCFs |
| --- | --- | --- |
| [Before](https://shard-qa.staging.cloud.databricks.com/#notebook/1444315/command/1444324) | 1.38m | 1.13 m |
| [After](https://shard-qa.staging.cloud.databricks.com/#notebook/1444320/command/1444321) | 1.68m | 1.25m |


core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -58,6 +62,40 @@ object BigVCFDatasource extends HlsUsageLogging {
...
+        VCFFileWriter.VCF_HEADER_KEY,
+        VCFFileWriter.INFER_HEADER) == VCFFileWriter.INFER_HEADER) {
+      // If we have to infer the header (explicitly or implicitly), determine all of the samples in the DataFrame.
+      // The samples that will be written consist of the distinct set of sample IDs identified across the DataFrame,
+      // in addition to missing sample IDs. The number of missing sample IDs will be equal to the maximum number
+      // found in any given row.
+      import data.sparkSession.implicits._
+      val distinctSampleLists =
+        data
+          .select("genotypes.sampleId")
```
henrydavidge: I think this will be an analysis exception if there is no `sampleId` genotype field.

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -58,6 +62,40 @@ object BigVCFDatasource extends HlsUsageLogging {
...
+          .as[Array[String]]
+          .distinct()
```
henrydavidge: Does `distinct` work correctly on array columns?

henrydavidge: It may be easier to reason about / more efficient to do
```
df.selectExpr("collect_set(explode(genotypes.sampleId))")`
```

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -58,6 +62,40 @@ object BigVCFDatasource extends HlsUsageLogging {
...
+          .collect
+      val maxNumMissingSamples = distinctSampleLists.map { sl =>
+        sl.count(_ == null)
+      }.max
+      val inferredSamples = distinctSampleLists
+        .flatten
+        .distinct
+        .dropWhile(_ == null)
+        .sorted
+      val maybeInferredSamples = if (inferredSamples.length > 0) {
```
henrydavidge: This should be an explicit argument to the stream writer, not passed through the option map.

core/src/main/scala/io/projectglow/vcf/VCFFileWriter.scala
```diff
@@ -73,7 +73,20 @@ object VCFFileWriter extends GlowLogging {
...
         logger.info("Inferring header for VCF writer")
-        (VCFSchemaInferrer.headerLinesFromSchema(schema).toSet, None)
+        val headerLines = VCFSchemaInferrer.headerLinesFromSchema(schema).toSet
+        if (options.contains("inferredSamples") || options.contains("numMissingSamples")) {
+          val inferredSamples =
+            options.get("inferredSamples").map(_.split("\t")).getOrElse(Array.empty)
+          val missingSamples = options
+            .get("numMissingSamples")
+            .map { n =>
+              Array.fill(n.toInt)("")
```
henrydavidge: Is this the same as the default sample ids that we inject in `VCFStreamWriter`?

henrydavidge: Also, it seems a little sketchy to me to inject sample ids if we don't have any indication that each row represents the same cohort. If each row has the same number of samples it makes sense to me, otherwise I think this is likely to return incomprehensible results.

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
karenfeng: `distinct` does work correctly on array columns. If we explode the sample IDs, I don't believe we'll be able to know how many sample IDs we're missing.

karenfeng: Also once we explode, there will be a single sample ID per row so there's not much use in calling `collect_set` on the `String`; I assume you meant `sampleIdDf.distinct()`.

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -100,4 +117,64 @@ object BigVCFDatasource extends HlsUsageLogging {
...
+    import data.sparkSession.implicits._
+    val hasSampleIds = genotypeSchemaOpt.get.exists(_.name == "sampleId")
+    if (hasSampleIds) {
+      val distinctSampleLists = data
+        .select("genotypes.sampleId")
+        .distinct()
+        .as[Array[String]]
+        .collect
+
+      val numMissingSampleList = distinctSampleLists.map { sl =>
```
henrydavidge: This behavior stills seems sort of complicated to me without a clear use case. Why would you ever have some sample ids but not others? I'm in favor of just writing the non-null sample ids.

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -100,4 +117,64 @@ object BigVCFDatasource extends HlsUsageLogging {
...
+      getMissingSamples(numMissingSampleList)
+    }
+  }
+
+  def getMissingSamples(numMissingSamplesList: Seq[Int]): Seq[String] = {
+    if (numMissingSamplesList.length > 1) {
+      throw new IllegalArgumentException(
+        "Rows contain varying number of missing samples; cannot infer VCF header.")
+    }
+    Array.fill(numMissingSamplesList.head)("")
```
henrydavidge: Why do we return empty strings instead of `sample_n`?

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
```diff
@@ -66,25 +64,74 @@ class VCFStreamWriter(
...
-    val maybeReplacedVC = maybeReplaceDefaultSampleIds(vc)
-    val vcBuilder = new VariantContextBuilder(maybeReplacedVC)
-    val iterator = maybeReplacedVC.getAttributes.entrySet().iterator()
-    while (iterator.hasNext) { // parse to string, then write,
-      // otherwise the write messes up double precisions
-      val entry = iterator.next()
-      vcBuilder.attribute(
-        entry.getKey,
-        VariantContextToVCFRowConverter.parseObjectAsString(entry.getValue))
+    if (replaceSampleIds) {
```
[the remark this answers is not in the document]
henrydavidge: Also, I think that this could all be simplified by splitting into two steps:
1. If we're replacing sample IDs, build a new genotype array.
2. If we've inferred sample IDs, iterate through the maybe rebuilt genotype array and verify that the sample ids match what we inferred.

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
```diff
@@ -93,69 +140,17 @@ class VCFStreamWriter(
...
     // Header must be written before closing writer, or else VCF readers will break.
-    if (!headerHasBeenSetOrWritten && writeHeader) {
-      if (providedSampleIds.isEmpty) {
+    if (header == null && writeHeader) {
+      if (inferSampleIds) {
         throw new IllegalStateException(
           "Cannot infer header for empty partition; " +
           "we suggest calling coalesce or repartition to remove empty partitions.")
       }
+      val (sampleIds, _) = sampleIdsMissingOpt.get
```
henrydavidge: Shouldn't be using the inferred sample IDs if we inferred them?

core/src/main/scala/io/projectglow/vcf/VCFWriterUtils.scala
```diff
@@ -0,0 +1,92 @@
...
+   *     - Missing, the sample IDs are injected from the number of genotypes per row (must be the same per row).
+   *     - Present, the sample IDs are found by unifying sample IDs across all rows.
+   */
+  def inferSampleIdsAndInjectMissing(data: DataFrame): (Seq[String], Boolean) = {
+    val genotypeSchemaOpt = data
+      .schema
+      .find(_.name == "genotypes")
+      .map(_.dataType.asInstanceOf[ArrayType].elementType.asInstanceOf[StructType])
+    if (genotypeSchemaOpt.isEmpty) {
+      logger.warn("No genotypes column, no sample IDs will be inferred.")
```
henrydavidge: Doesn't seem like this should be a warning. Probably just an INFO or even DEBUG.

core/src/main/scala/io/projectglow/vcf/VCFWriterUtils.scala
```diff
@@ -0,0 +1,92 @@
...
+      return (Seq.empty, false)
+    }
+    val genotypeSchema = genotypeSchemaOpt.get
+
+    import data.sparkSession.implicits._
+    val hasSampleIdsColumn = genotypeSchema.exists(_.name == "sampleId")
```
henrydavidge: This logic is a little convoluted. I think it would be more clear to split into two steps:
1. Collect all distinct sample ids. If there are no missings, we're done. If there are missings and defined sample ids, throw. If there are only missings, proceed to step 2.
2. Check that every genotype row has the same length.

core/src/test/scala/io/projectglow/vcf/VCFFileWriterSuite.scala
```diff
@@ -515,70 +626,117 @@ class SingleFileVCFWriterSuite extends VCFFileWriterSuite("bigvcf") {
...
+      spark
+        .read
+        .format(readSourceName)
+        .option("vcfRowSchema", "true")
+        .load(tempFile)
+
+    val sess = spark
+    import sess.implicits._
+
+    val sampleIdRows = rereadDf.select("genotypes.sampleId").distinct().as[Seq[String]].collect
```
henrydavidge: add comments, this is complicated

docs/source/etl/variant-data.rst
```diff
@@ -118,6 +118,14 @@ For both the single and sharded VCF writer, you can use the following option to
 |                 |        |             | this file is used as the VCF header for each partition.                                                            |
 +-----------------+--------+-------------+--------------------------------------------------------------------------------------------------------------------+
 
+.. _infer-vcf-samples:
+
+If the header is inferred from the DataFrame, the sample IDs are derived from the rows. If there are missing sample IDs,
+they will be represented as ``sample_n``, for which ``n`` reflects the index of the sample in a row.
+
+- For the single VCF writer, the inferred sample IDs are the distinct set of all sample IDs from the DataFrame.
```
henrydavidge: let's call it the big vcf writer so there's only one term

docs/source/etl/variant-data.rst
```diff
@@ -118,6 +118,14 @@ For both the single and sharded VCF writer, you can use the following option to
...
+- For the sharded VCF writer, the sample IDs are inferred from the first row of each partition. If the rows do not
```
henrydavidge: say that it's an error if every row doesn't have the same sample ids or the same number of samples if there are no sample ids

docs/source/etl/variant-data.rst
```diff
@@ -118,6 +118,14 @@ For both the single and sharded VCF writer, you can use the following option to
 |                 |        |             | this file is used as the VCF header for each partition.                                                            |
 +-----------------+--------+-------------+--------------------------------------------------------------------------------------------------------------------+
 
+.. _infer-vcf-samples:
+
+If the header is inferred from the DataFrame, the sample IDs are derived from the rows. If there are missing sample IDs,
+they will be represented as ``sample_n``, for which ``n`` reflects the index of the sample in a row.
```
henrydavidge: say that in this case there must be the same number of samples in each row.

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
```diff
@@ -66,25 +64,74 @@ class VCFStreamWriter(
...
-    while (iterator.hasNext) { // parse to string, then write,
-      // otherwise the write messes up double precisions
-      val entry = iterator.next()
-      vcBuilder.attribute(
-        entry.getKey,
-        VariantContextToVCFRowConverter.parseObjectAsString(entry.getValue))
+    if (replaceSampleIds) {
+      val oldGts = vcBuilder.getGenotypes
+      if (inferSampleIds && oldGts.size != header.getNGenotypeSamples) {
+        if (oldGts.asScala.map(_.getSampleName).exists(s => s != "")) {
```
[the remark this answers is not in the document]
karenfeng: Actually, not quite... This happens inside the case that the number of samples don't match.

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
karenfeng: This happens in the case that `header == null`, which only occurs in the case that we did not encounter any rows to write and therefore cannot have inferred any sample IDs.

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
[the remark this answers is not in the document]
karenfeng: I believe that if 1 is true, then we don't need 2.

core/src/main/scala/io/projectglow/vcf/VCFWriterUtils.scala
```diff
@@ -0,0 +1,107 @@
...
+
+    if (hasSampleIdsColumn) {
+      val distinctSampleIds = data
+        .selectExpr("explode(genotypes.sampleId)")
+        .distinct()
+        .as[String]
+        .collect
+      val numPresentSampleIds = distinctSampleIds.count(!sampleIsMissing(_))
```
henrydavidge: Do these need to be sorted?

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
```diff
@@ -65,20 +63,62 @@ class VCFStreamWriter(
...
+    replaceSampleIds = samplesIdsFromMissing.fromMissing
+  }
 
-    val maybeReplacedVC = maybeReplaceDefaultSampleIds(vc)
-    val vcBuilder = new VariantContextBuilder(maybeReplacedVC)
-    val iterator = maybeReplacedVC.getAttributes.entrySet().iterator()
+  // Replace genotypes' missing sample IDs with those from the header
+  def replaceMissingSampleIds(vcBuilder: VariantContextBuilder): VariantContextBuilder = {
+    val oldGts = vcBuilder.getGenotypes
+    if (inferSampleIds && oldGts.size != header.getNGenotypeSamples) {
```
henrydavidge: I think we could just do this check in `checkInferredSampleIds`

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
```diff
@@ -65,20 +63,62 @@ class VCFStreamWriter(
...
+      VCFWriterUtils.throwSampleInferenceFailure()
+    }
+    val newGts = new JArrayList[Genotype](oldGts.size)
+    var i = 0
+    while (i < oldGts.size) {
+      val oldGt = oldGts.get(i)
+      if (inferSampleIds && !oldGt.getSampleName.isEmpty) {
```
henrydavidge: Might be more understandable to do this check in `checkInferredSampleIds`. Here we just replace all non-empty sample ids.

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
karenfeng: `checkInferredSampleIds` checks whether the sample IDs are a subset of those in the header; this length check needs to happen ASAP or else we'll run into an index out of bounds exception.

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
karenfeng: By non-empty do you mean empty?

karenfeng: I also think it's more understandable to catch this upfront, before we try replacing any non-missing sample IDs.

core/src/main/scala/io/projectglow/vcf/VCFWriterUtils.scala
karenfeng: These will be sorted later. There's no real need for them to be sorted, but it's a one-time calculation and it'd probably be good for the VCF's sample IDs to be non-random.

core/src/main/scala/io/projectglow/vcf/BigVCFDatasource.scala
```diff
@@ -59,12 +59,24 @@ object BigVCFDatasource extends HlsUsageLogging {
...
+        .forall(_ == VCFHeaderUtils.INFER_HEADER)) {
       throw new SparkException("Cannot infer header for empty VCF.")
     }
+
+    val (headerLineSet, providedSampleIdsOpt) = VCFHeaderUtils.parseHeaderLinesAndSamples(
+      options,
+      Some(VCFHeaderUtils.INFER_HEADER),
+      schema,
+      conf)
+    val sampleIdsFromMissing = if (providedSampleIdsOpt.isDefined) {
```
henrydavidge: nit: can we call this `sampleIdInfo`?

core/src/main/scala/io/projectglow/vcf/VCFHeaderUtils.scala
```diff
@@ -0,0 +1,87 @@
...
+  def parseHeaderFromString(s: String): VCFHeader = {
+    val stringReader = new StringReader(s)
+    val lineIterator = new LineIteratorImpl(IOUtils.lineIterator(stringReader).asScala)
+    val codec = new VCFCodec()
+    try {
+      codec.readActualHeader(lineIterator).asInstanceOf[VCFHeader]
+    } catch {
+      case NonFatal(e) =>
+        logger.warn(s"Wasn't able to parse VCF header in $s. $e")
+        new VCFHeader()
```
henrydavidge: I can't imagine that this is the correct error behavior... shouldn't we throw this exception?

core/src/main/scala/io/projectglow/vcf/VCFInputFormatter.scala
```diff
@@ -40,10 +40,15 @@ class VCFInputFormatter(
 
   override def init(stream: OutputStream): Unit = {
     this.stream = stream
+    val sampleIdsMissingOpt = if (providedSampleIdsOpt.isDefined) {
```
henrydavidge: `sampleIdInfo`

core/src/main/scala/io/projectglow/vcf/VCFStreamWriter.scala
```diff
@@ -17,46 +17,43 @@
...
+ * If missing sample IDs were used to set the header, sample IDs in all rows to be written will be replaced with those
+ * from the header.
+ * If all sample IDs are present when setting the header, sample IDs in written rows will not be replaced.
+ * Mixed missing/present sample IDs are not permitted.
  *
- * If default header sample IDs were used, any written variant context's empty sample IDs are also
- * replaced with defaults.
+ * @param stream The stream to write to
+ * @param headerLineSet Header lines used to set the VCF header
+ * @param sampleIdInfoOpt Sample IDs (and if any were missing) if pre-determined, None otherwise
```
henrydavidge: and if any were missing -> and whether they were injected because the ids were missing

core/src/main/scala/io/projectglow/vcf/VCFHeaderUtils.scala
karenfeng: Huh, I don't remember why we did this. I'll change it.

Issue #86, opened 2019-11-25
Struct transformations, add_struct_fields example fails.


brechtmann:
Running the add_struct fields example in the documentation seems to work first, but when calling `.show()` the example fails.

Setting up the initial DataFrame:
```
from pyspark.sql import Row
row_one = Row(Row(str_col='foo', int_col=1, bool_col=True))
row_two = Row(Row(str_col='bar', int_col=2, bool_col=False))
base_df = spark.createDataFrame([row_one, row_two], schema=['base_col'])
base_df.selectExpr("subset_struct(base_col, 'str_col', 'bool_col') as subsetted_col").show()
```
```
added_col_df = base_df.selectExpr("add_struct_fields(base_col, 'float_col', 3.14, 'rev_str_col', \
    reverse(base_col.str_col)) as added_col")
```
```
added_col_df.show()
```
Calling .show() gives me the following error:
```
---------------------------------------------------------------------------
Py4JJavaError                             Traceback (most recent call last)
<ipython-input-37-1fafef989fbe> in <module>
----> 1 added_col_df.show()

/opt/modules/i12g/anaconda/3-2018.12/envs/glow-test/lib/python3.7/site-packages/pyspark/sql/dataframe.py in show(self, n, truncate, vertical)
    378         """
    379         if isinstance(truncate, bool) and truncate:
--> 380             print(self._jdf.showString(n, 20, vertical))
    381         else:
    382             print(self._jdf.showString(n, int(truncate), vertical))

/opt/modules/i12g/anaconda/3-2018.12/envs/glow-test/lib/python3.7/site-packages/py4j/java_gateway.py in __call__(self, *args)
   1255         answer = self.gateway_client.send_command(command)
   1256         return_value = get_return_value(
-> 1257             answer, self.gateway_client, self.target_id, self.name)
   1258 
   1259         for temp_arg in temp_args:

/opt/modules/i12g/anaconda/3-2018.12/envs/glow-test/lib/python3.7/site-packages/pyspark/sql/utils.py in deco(*a, **kw)
     61     def deco(*a, **kw):
     62         try:
---> 63             return f(*a, **kw)
     64         except py4j.protocol.Py4JJavaError as e:
     65             s = e.java_exception.toString()

/opt/modules/i12g/anaconda/3-2018.12/envs/glow-test/lib/python3.7/site-packages/py4j/protocol.py in get_return_value(answer, gateway_client, target_id, name)
    326                 raise Py4JJavaError(
[...]

PR #87, opened 2019-11-25
Inject Spark session rules during SQL extension registration

karenfeng:

When Glow is attached as a library (vs. being baked into a Spark runtime), we need to inject the rules in addition to registering the extensions. As the rules are needed for the extensions to function properly, it makes sense for simplicity's sake to perform the injection during the `glow.register(spark)` step. This PR also exposes the same `Glow.register(spark)` API in Scala, and simplifies documentation for Scala users by creating distinct Python/Scala tabs for language-specific docs.

Resolves https://github.com/projectglow/glow/issues/86.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

Replicated the attached issue in a Jupyter notebook. After running the following step in a cell:
```
spark.sparkContext._jvm.io.projectglow.sql.GlowSQLExtensions().apply(spark._jsparkSession.extensions())
```
The command succeeded.

### DIFF
--- a/core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
+++ b/core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
@@ -237,12 +237,12 @@ object VCFFileFormat {
       } else {
         new PositionalBufferedStream(is)
       }
-
-      val vcfCodec = new VCFCodec()
-      val reader = new AsciiLineReaderIterator(AsciiLineReader.from(wrappedStream))
-      val header = vcfCodec
-        .readActualHeader(reader)
-      (header.asInstanceOf[VCFHeader], vcfCodec)
+      WithUtils.withCloseable(wrappedStream) { ws =>
+        val vcfCodec = new VCFCodec()
+        val reader = new AsciiLineReaderIterator(AsciiLineReader.from(ws))
+        val header = vcfCodec.readActualHeader(reader)
+        (header.asInstanceOf[VCFHeader], vcfCodec)
+      }
     }
   }
 
--- a/core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
+++ b/core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
@@ -237,12 +237,10 @@ object VCFFileFormat {
       } else {
         new PositionalBufferedStream(is)
       }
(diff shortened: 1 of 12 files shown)


python/glow/glow.py
```diff
@@ -37,9 +37,10 @@ def register(session: SparkSession):
...
     register(session)
 
-    Register SQL extensions for a Spark session.
+    Register SQL extensions and inject the relevant rules for a Spark session.
 
     :param session: Spark session
     """
     assert check_argument_types()
     session._jvm.io.projectglow.sql.SqlExtensionProvider.register(session._jsparkSession)
+    session._jvm.io.projectglow.sql.GlowSQLExtensions().apply(session._jsparkSession.extensions())
```
henrydavidge: Shall we do this in `SqlExtensionProvider.register`?

core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
```diff
@@ -237,12 +237,12 @@ object VCFFileFormat {
       } else {
         new PositionalBufferedStream(is)
       }
-
-      val vcfCodec = new VCFCodec()
-      val reader = new AsciiLineReaderIterator(AsciiLineReader.from(wrappedStream))
-      val header = vcfCodec
-        .readActualHeader(reader)
-      (header.asInstanceOf[VCFHeader], vcfCodec)
+      WithUtils.withCloseable(wrappedStream) { ws =>
```
henrydavidge: Why this change?

docs/source/getting-started.rst
```diff
@@ -4,42 +4,68 @@ Getting Started
...
+        `download a specific distribution <https://spark.apache.org/downloads.html>`_.
+
+        Start the `Spark shell <http://spark.apache.org/docs/latest/rdd-programming-guide.html#using-the-shell>`_
+        with the Glow maven package:
+
+        .. substitution-code-block:: sh
+
+          ./bin/spark-shell --packages io.projectglow:glow_2.11:|mvn-version|
+
+        And now your notebook is glowing! To access the Glow functions, you need to register them with the
```
henrydavidge: nit: there is no notebook here so it can't be glowing :)

PR #88, opened 2019-11-25
Add permute array

LelandBarnard:
Added new sql expression permute_array, which is used to subset and/or reorder an array with a given array of indices

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

### DIFF
--- a/core/src/test/scala/io/projectglow/tertiary/PermuteArrayExprSuite.scala
+++ b/core/src/test/scala/io/projectglow/tertiary/PermuteArrayExprSuite.scala
@@ -1,9 +1,25 @@
+/*
+ * Copyright 2019 The Glow Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


core/src/test/scala/io/projectglow/tertiary/PermuteArrayExprSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+    val aRow = spark.range(1).select(lit(dataArray) as "dataArray", lit(indexArray) as "indexArray")
+    val permutedData = aRow
+      .selectExpr("permute_array(dataArray,indexArray) as permutedData")
+      .take(1)
+      .map(r => r.getAs[Seq[Int]]("permutedData"))
+      .head
+    assert(permutedData == indexArray.map(i => dataArray(i - 1)).toSeq)
+  }
+
+  test(testName = "null indices") {
```
henrydavidge: nit: you don't need `testName = ` for each of the test cases.

PR #89, opened 2019-11-26
Add Python doc tests

karenfeng:

Make sure the code blocks in our documentation work as expected.

## How is this patch tested?
- [ ] Unit tests
- [x] Integration tests
- [x] Manual tests

`sbt docs/test`


python/environment.yml
```diff
@@ -4,10 +4,12 @@ dependencies:
   - pytest
   - pip
   - pip:
+    - numpy
```
henrydavidge: we should have a version for this

build.sbt
```diff
@@ -177,6 +178,29 @@ lazy val python =
       publish / skip := true
     )
 
+lazy val docs =
+  (project in file("docs"))
+    .dependsOn(core % "test->test")
+    .settings(
+      test in Test := {
+        // Pass the test classpath to pyspark so that we run the same bits as the Scala tests
+        val classpath = (fullClasspath in Test)
```
henrydavidge: Can we factor this into `commonSettings` (or other shared value)? Since tis' used here and in `python`.

conftest.py
```diff
@@ -2,9 +2,10 @@
 import pytest
 
 
+# Shared across Python and docs tests
 @pytest.fixture(scope="session")
 def spark():
-    sess = SparkSession.builder\
+    sess = SparkSession.builder \
         .master("local[2]") \
         .config("spark.serializer", "org.apache.spark.serializer.KryoSerializer") \
```
henrydavidge: Can we remove this now?

karenfeng: As discussed offline, we still use Kryo serializers in all of our tests (even if we don't require registration).

PR #95, opened 2019-12-06
Performance improvements for linear and logistic regression test

henrydavidge:
Speeds up the regression kernels, especially logistic regression, by:
- Implementing codegen'd versions. In these versions, we store state in thread local variables that are cleared at the end of each task.
- In logistic regression, allocate memory needed for temporary variables in the fit state. These buffer vectors and matrices are then used in the iterative fitting process to avoid repeated allocations.
- In logistic regression, only deserialize the covariate matrix when we actually need it.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

Behavior does not change as a result of this PR, so existing unit tests suffice for correctness.

On ~1m variants and 500k samples, the runtime changes are:
- LRT logistic regression: 23.2hr -> 9.4hr
- Linear regression: 8.3hr -> 4.4hr

For logistic regression, the runtime was 16.2hr after minimizing allocations and copying. The rest of the difference and all of the difference for linear regression comes from implementing codegen.


core/src/main/scala/io/projectglow/sql/expressions/LikelihoodRatioTest.scala
```diff
@@ -22,30 +22,52 @@ import org.apache.spark.sql.Encoders
...
 
-    if (!nullFit.converged || !fullFit.converged) {
+    val fullFit =
+      LogisticRegressionGwas.newtonIterations(
+        fitState.x,
+        phenotypes,
+        fitState.hessian,
+        fitState.newtonState)
+
+    if (!fitState.nullFit.converged || !fullFit.converged) {
```
karenfeng: We can probably return early if the null fit explodes.

core/src/main/scala/io/projectglow/sql/expressions/LogisticRegressionExpr.scala
```diff
@@ -19,29 +19,102 @@ package io.projectglow.sql.expressions
...
 
+class LogisticRegressionState(testStr: String) {
+  val logitTest = LogisticRegressionGwas
+    .logitTests
+    .getOrElse(
+      testStr,
+      throw new IllegalArgumentException(
+        s"Supported tests are currently: ${LogisticRegressionGwas.logitTests.keys.mkString(", ")}")
+    )
+  var fitState: logitTest.FitState = _
```
karenfeng: Comment that only one of `fitState` or `nullFitMap` should be non-`null`/empty respectively.

core/src/main/scala/io/projectglow/sql/expressions/LinearRegressionExpr.scala
```diff
@@ -16,19 +16,36 @@
...
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
+import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, CodegenFallback, ExprCode}
 import org.apache.spark.sql.catalyst.expressions.{Expression, ImplicitCastInputTypes, TernaryExpression}
 import org.apache.spark.sql.catalyst.util.ArrayData
 import org.apache.spark.sql.types._
 
+object LinearRegressionExpr {
+  val matrixUDT = SQLUtils.newMatrixUDT()
+  val state = new ThreadLocal[CovariateQRContext]
```
karenfeng: Can we make an abstraction for how we get/set/remove state? The general logic seems to match between logit and linreg.

core/src/main/scala/io/projectglow/sql/expressions/LogisticRegressionGwas.scala
```diff
@@ -175,27 +179,33 @@ case class NewtonResult(
...
-    throw new IllegalArgumentException("Test does not support null model fit")
-  }
-}
+  /**
+   * Initializes a [[FitState]] for a (covariate matrix, phenotype array) pair.
+   *
+   * As much memory allocation as possible should be performed in this step to avoid allocations
+   * in the per-row fit.
+   */
+  def init(phenotypes: Array[Double], covariates: SparkDenseMatrix): FitState
```
karenfeng: If fit state per phenotype is false, then `phenotypes` will not be used in the initialization, correct?

core/src/main/scala/io/projectglow/sql/expressions/FirthTest.scala
```diff
@@ -19,83 +19,93 @@ package io.projectglow.sql.expressions
...
-    b(0 until m0) := nullFit.b
-    val fullFit = fitFirth(x, y, b)
+    fitState.x(::, -1) := genotypes
+    fitState.nullFitArgs.clear()
+    val nullFit = fitFirth(fitState.x, phenotypes, fitState.nullFitArgs)
+    fitState.fullFitArgs.clear()
+    fitState.fullFitArgs.initFromNullFit(nullFit.fitState)
+    val fullFit = fitFirth(fitState.x, phenotypes, fitState.fullFitArgs)
 
     if (!nullFit.converged || !fullFit.converged) {
```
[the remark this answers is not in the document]
henrydavidge: Good point

core/src/main/scala/io/projectglow/sql/expressions/LinearRegressionExpr.scala
henrydavidge: I tried to do that, but I don't think there's actually much to share other than registering the callback.

core/src/main/scala/io/projectglow/sql/expressions/LogisticRegressionGwas.scala
henrydavidge: True

Issue #97, opened 2019-12-06
Issue with Glow liftOver function


habeggel:
Overall, the Glow liftOver function seems to work as expected, but there is an issue when the REF/ALT alleles are swapped.  Below is an example:
 
VCF entry:
`22   16849933   .    C    T    .    .    .`
 
Output from running Picard LiftoverVcf (see details below):
`22   16369271   .    T    C    .    PASS     OriginalContig=22;OriginalStart=16849933;SwappedAlleles`
 
Output after running the Glow liftOver:
 
`val df_liftOver = Glow.transform("lift_over_variants", df_b37, Map("reference_file" -> referencePath, "chain_file" -> chain)).drop("genotypes")`
 
<img width="1474" alt="Screen Shot 2019-12-06 at 11 31 33 AM" src="https://user-images.githubusercontent.com/6634418/70353603-bcece400-183b-11ea-9b63-9f1b6b6f0a66.png">

 
It appears that the Glow liftOver fails due to a “MismatchedRefAllele”.  
 
However, the desired behavior would be as follows:

- Liftover the site to the new genomic coordinates (`22:16369271`)
- Swap the reference and alternate alleles
- Set INFO_SwappedAlleles to ‘true’

PR #100, opened 2019-12-10
Support swapping ref/alt alleles during liftover

karenfeng:

Changes the default behavior of the [liftover transformer](https://glow.readthedocs.io/en/latest/etl/lift-over.html#variant-liftover) to swap reference and alternate alleles. This is equivalent to setting the option `RECOVER_SWAPPED_REF_ALT=TRUE` for [Picard's LiftoverVcf](https://software.broadinstitute.org/gatk/documentation/tooldocs/4.0.11.0/picard_vcf_LiftoverVcf.php).

Resolves https://github.com/projectglow/glow/issues/97.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


core/src/main/scala/io/projectglow/transformers/LiftOverVariantsTransformer.scala
```diff
@@ -246,6 +246,15 @@ object LiftOverVariantsTransformer extends GlowLogging {
     val refStr = refSeq.getBaseString.substring(outputVc.getStart - 1, outputVc.getEnd)
 
     if (outputVc.getReference.getBaseString.toLowerCase != refStr.toLowerCase) {
+      if (outputVc.isBiallelic && outputVc.isSNP && refStr.equalsIgnoreCase(
```
henrydavidge: Why is this only for biallelic variants?

karenfeng: This logic is pulled from `LiftoverVcf`, which only recovers swapping ref/alt alleles in biallelic SNPs: https://github.com/broadinstitute/picard/blob/a3021a7a7e00fcd5f05a9d6088c8f29a707d7420/src/main/java/picard/vcf/LiftoverVcf.java#L541

PR #103, opened 2019-12-10
[WIP] Accelerate VariantContexts to InternalRow conversion

karenfeng:

To create a `GenericArrayData` cheaply, we perform a series of optimizations:
- Use `Array`s instead of mutable Scala collections
- Use `Array.copy` instead of `arr.take` to avoid creating an `ArrayBuilder`
- Use `Array[Any]` in order to directly create a `GenericArrayData`

To create the converter more cheaply, we replace `schema.map` with an `Array`.

We also replace the `RowConverter.Updater`s with classes that implement `apply` rather than depending on anonymous functions.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


karenfeng:
I am closing this PR as these speedups are primarily for `VCFRow`s, which we plan to deprecate.

PR #112, opened 2019-12-17
Add option to specify bgzf

henrydavidge:

Inferring whether input files are bgzf can be quite slow when there are many files because we need to open a file stream. There's not a straightforward way to parallelize this in the file format interface, so we can add an option to force bgzf.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


henrydavidge:
Closing in favor of #114

PR #114, opened 2019-12-18
Move gzip check to file reading

henrydavidge:

Previously we checked whether a .vcf.gz file was block gzipped in `VCFFileFormat.isSplittable`. Although this location is logically the right place, Spark runs this code on each file serially. Since checking whether a file is gzipped or block gzipped requires reading the part of the file, this process can be very slow on large datasets.

Instead, we now handle the difference between gz and bgz while reading files. If we detect that a file is gzipped, we read the entire file if the current split starts at the beginning of the file and read nothing if it starts in the middle. Note that this is not the ideal behavior since we're creating a ton of tasks that do nothing. However, in the common case where .gz actually means block gzip, we can now skip the expensive serial check without requiring more user input.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
+++ b/core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
@@ -68,12 +68,7 @@ class VCFFileFormat extends TextBasedFileFormat with DataSourceRegister with Hls
       )
     }
 
-    codecFactory.getCodec(path) match {
-      case null => true
-      case _: BGZFEnhancedGzipCodec =>
-        VCFFileFormat.isValidBGZ(path, sparkSession.sparkContext.hadoopConfiguration)
-      case c => c.isInstanceOf[SplittableCompressionCodec]
-    }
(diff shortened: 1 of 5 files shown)


core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
```diff
@@ -68,12 +68,7 @@ class VCFFileFormat extends TextBasedFileFormat with DataSourceRegister with Hls
       )
     }
 
-    codecFactory.getCodec(path) match {
-      case null => true
-      case _: BGZFEnhancedGzipCodec =>
-        VCFFileFormat.isValidBGZ(path, sparkSession.sparkContext.hadoopConfiguration)
-      case c => c.isInstanceOf[SplittableCompressionCodec]
-    }
+    codecFactory.getCodec(path).isInstanceOf[SplittableCompressionCodec]
```
karenfeng: Can you add a comment saying that we'll lazily determine whether this is bgzipped later?

Issue #113, opened 2019-12-18
liftOver issue: adjusting the genotypes array when alleles are swapped


habeggel:
I tested Glow to liftOver a GRCh37 BGEN file to GRCh38.  Ideally, sites with swapped alleles should get corrected and the genotypes array should get adjusted accordingly.  However, it appears that the “posteriorProbabilities” in the genotypes array are not swapped for sites where “INFO_SwappedAlleles” is true.

References: 
- Issue: https://github.com/projectglow/glow/issues/97
- Pull request: https://github.com/projectglow/glow/pull/100

PR #118, opened 2019-12-23
Reverse arrays during liftover if ref/alt alleles swapped

karenfeng:

In the case that ref/alt alleles are swapped by the variant liftover transformer, the Picard implementation (that we currently use) performs the following changes:
- For the VariantContext
   + Swap alleles
   + Set AF to 1-AF
   + Drop MAX_AF
- For each Genotype
   + Change calls to match the new alleles
   + Reverse AD
   + Reverse PL

All other fields are untouched. The Picard implementation does allow other single-value INFO fields (say, X) to be set to 1-X or dropped, but not to reverse arrays.

This implementation matches the previous behavior, but also finds INFO and FORMAT array-based fields and reverses them.

Addresses https://github.com/projectglow/glow/issues/113.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


core/src/main/scala/io/projectglow/transformers/LiftOverVariantsTransformer.scala
```diff
@@ -206,7 +215,9 @@ object LiftOverVariantsTransformer extends GlowLogging {
       inputVc: VariantContext,
       liftOver: LiftOver,
       minMatchRatio: Double,
-      refSeqMap: Map[String, ReferenceSequence]): (Option[VariantContext], Option[String]) = {
+      refSeqMap: Map[String, ReferenceSequence],
+      attributesToSwap: Seq[String],
```
henrydavidge: Maybe it would be more clear to call this `infoFieldsToSwap`? And `formatFieldsToSwap` or `genotypeFieldsToSwap`? I don't think it's helpful to use the htsjdk terminology as it's pretty bad.

core/src/main/scala/io/projectglow/vcf/InternalRowToVariantContextConverter.scala
```diff
@@ -457,22 +458,21 @@ class InternalRowToVariantContextConverter(
...
-    val base = VariantContextToVCFRowConverter.parseObjectAsString(valueToConvert)
-    if (base.isEmpty) {
-      // Missing values are represented by '.' instead of empty strings
-      VCFConstants.MISSING_VALUE_v4
-    } else {
-      base
+    value match {
+      case null => VCFConstants.MISSING_VALUE_v4
+      case "" => VCFConstants.MISSING_VALUE_v4
+      case _: JBoolean => VCFConstants.MISSING_VALUE_v4
```
henrydavidge: Is this correct?

core/src/main/scala/io/projectglow/transformers/LiftOverVariantsTransformer.scala
```diff
@@ -280,4 +366,51 @@ object LiftOverVariantsTransformer extends GlowLogging {
...
+  def getExtendedAttributesToSwap(variantSchema: StructType): Seq[String] = {
+    val genotypeFieldOpt = variantSchema
+      .fields
+      .find(_.name == VariantSchemas.genotypesFieldName)
+
+    if (genotypeFieldOpt.isEmpty) {
+      return Seq.empty
+    }
+
+    val genotypeField =
```
henrydavidge: Could we throw a better exception if `genotypes` is not an `ArrayType(StructType)`? It would be nice to put this in a common place since we frequently need to get the genotype schema.

PR #122, opened 2020-01-02
Parse VCF variant annotations

karenfeng:

Parses CSQ and ANN variant annotation fields (usually set by VEP and SnpEff, respectively) to simplify querying; right now, users must use fairly complex Spark SQL expressions to access fields of interest.

### Usage example

Previously, finding the LoF field for variants output by VEP required the following, where `23` is hard-coded based on the position of the `LoF` string in the header line: `vep_df.selectExpr(expr("transform(INFO_CSQ, c -> split(c, '[|]')[23])"))`.
Now, this is as simple as `vep_df.selectExpr("INFO_CSQ.LoF")`

### Schema
This PR sets the schema for the variant annotation fields to be an array of structs, whose field names are defined by the header line and whose field types default to strings (except for special cases). For example, the schema of a VCF annotated by SnpEff would be as follows:
```
 |-- INFO_ANN: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- Allele: string (nullable = true)
 |    |    |-- Annotation: array (nullable = true)
 |    |    |    |-- element: string (containsNull = true)
 |    |    |-- Annotation_Impact: string (nullable = true)
 |    |    |-- Gene_Name: string (nullable = true)
 |    |    |-- Gene_ID: string (nullable = true)
 |    |    |-- Feature_Type: string (nullable = true)
 |    |    |-- Feature_ID: string (nullable = true)
 |    |    |-- Transcript_BioType: string (nullable = true)
 |    |    |-- Rank: struct (nullable = true)
 |    |    |    |-- rank: integer (nullable = true)
 |    |    |    |-- total: integer (nullable = true)
 |    |    |-- HGVS.c: string (nullable = true)
 |    |    |-- HGVS.p: string (nullable = true)
 |    |    |-- cDNA.pos / cDNA.length: struct (nullable = true)
 |    |    |    |-- pos: integer (nullable = true)
 |    |    |    |-- length: integer (nullable = true)
 |    |    |-- CDS.pos / CDS.length: struct (nullable = true)
 |    |    |    |-- pos: integer (nullable = true)
 |    |    |    |-- length: integer (nullable = true)
 |    |    |-- AA.pos / AA.length: struct (nullable = true)
 |    |    |    |-- pos: integer (nullable = true)
 |    |    |    |-- length: integer (nullable = true)
 |    |    |-- Distance: integer (nullable = true)
 |    |    |-- ERRORS / WARNINGS / INFO: string (nullable = true)
```

#### Assumptions

This PR assumes that the header line defines the field names in the description as a series of pipe (`|`)-delimited strings following a colon (`:`). I stripped single-quotes from the description and trimmed white-space from the field names as well.

- VEP header line: `##INFO=<ID=CSQ,Number=.,Type=String,Description="Consequence annotations from Ensembl VEP. Format: Allele|Consequence|IMPACT|SYMBOL|Gene|Feature_type|Feature|BIOTYPE|EXON|INTRON|HGVSc|HGVSp|cDNA_position|CDS_position|Protein_position">`
- SnpEff header line: `##INFO=<ID=ANN,Number=.,Type=String,Description="Functional annotations: 'Allele | Annotation | Annotation_Impact | Gene_Name | Gene_ID | Feature_Type | Feature_ID | Transcript_BioType | Rank | HGVS.c | HGVS.p | cDNA.pos / cDNA.length | CDS.pos / CDS.length | AA.pos / AA.length | Distance | ERRORS / WARNINGS / INFO' ">`

#### Design

This PR creates a shared annotation parser for ANN and CSQ. The schema for any annotation subfield defaults to a string, but there are special cases for subfields seen in SnpEff and vanilla VEP (+Loftee). There are more possible subfields if the user uses VEP plugins, but we can extend to those based on demand.

This backend may also be used to support parsing other annotation fields in the future (eg. EFF, NMF and LOF). I did not implement EFF, NMF and LOF parsers in this PR.
- This is partly for simplicity, as EFF is not the default SnpEff output field as of v4.1+. The NMF and LOF fields are also comparatively simple, with only 4 sub-fields each.
- This is partly also because these fields do not match the standard annotation format. More specifically, they have a more complex notation; rather than delimiting all sub-fields with the pipe character, they also use parentheses.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


henrydavidge:
Why are the annotations of type `ArrayType(StructType)`? I think that some particular annotations may have repeated values, but most do not. Also, perhaps we could provide better types for the most common annotations (like those in default VEP and loftee) and provide other annotations as strings.

karenfeng:
> Why are the annotations of type `ArrayType(StructType)`? I think that some particular annotations may have repeated values, but most do not. Also, perhaps we could provide better types for the most common annotations (like those in default VEP and loftee) and provide other annotations as strings.

As expressed in the header line, the number of annotations in any given VCF row can vary. The VEP docs state that: Each prediction, for a given variant, is separated by the character "," in the CSQ INFO field (e.g. when a variant overlaps more than 1 transcript).

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
```diff
@@ -285,6 +286,48 @@ class VariantContextToInternalRowConverter(
...
+      schema: StructType): Array[GenericInternalRow] = {
+    vc.getAttributeAsStringList(key, "")
+      .asScala
+      .toArray
+      .map { effect =>
+        // Providing a limit to the splitter preserves empty annotations
+        val stringAnnotations =
+          effect.split(AnnotationUtils.annotationDelimiterRegex, schema.size)
+        val convertedAnnotations = stringAnnotations.zip(schema.fields).map {
+          case ("", _) => null
```
henrydavidge: What's up with this case?

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
```diff
@@ -285,6 +286,48 @@ class VariantContextToInternalRowConverter(
...
+      vc: VariantContext,
+      key: String,
+      schema: StructType): Array[GenericInternalRow] = {
+    vc.getAttributeAsStringList(key, "")
+      .asScala
+      .toArray
+      .map { effect =>
+        // Providing a limit to the splitter preserves empty annotations
+        val stringAnnotations =
+          effect.split(AnnotationUtils.annotationDelimiterRegex, schema.size)
```
henrydavidge: Do you know if `String.split` actually uses regular expressions for one character patterns like this? If so we may be better off using Guava's `Splitter`.

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
```diff
@@ -285,6 +286,48 @@ class VariantContextToInternalRowConverter(
...
+        val convertedAnnotations = stringAnnotations.zip(schema.fields).map {
+          case ("", _) => null
+          case (strAnn, annField) =>
+            annField.dataType match {
+              case ArrayType(StringType, _) =>
+                new GenericArrayData(
+                  strAnn
+                    .split(AnnotationUtils.arrayDelimiter)
+                    .map(UTF8String.fromString(_).asInstanceOf[Any]))
+              case st if st.isInstanceOf[StructType] =>
```
henrydavidge: This case is not obvious. It's for the multi field annotations? Please document.

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
```diff
@@ -285,6 +286,48 @@ class VariantContextToInternalRowConverter(
     makeArray(strList, parseFn)
   }
 
+  private def getAnnotationArray(
```
henrydavidge: Can we write this in Java or use performant scala style? Whenever we use scala collections like this we end up rewriting for performance.

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
karenfeng: It does, at least since Java 7 - see http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/String.java#l2316.

PR #124, opened 2020-01-07
Save INFO/FORMAT multi-value attributes as ArrayLists

karenfeng:

Currently we save multi-value INFO/FORMAT attributes as `Array[AnyRef]`s in a `VariantContext`. This can break libraries that modify `VariantContext`s, as their multi-value attributes are commonly interpreted as lists (`CommonInfo.getAttributeAsList`). For example, the GATK's `StrandBiasTest` expects either a `String` or `ArrayList[Integer]`.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


henrydavidge:
I'll defer to @kianfar77 for a review as I remember he worked through a similar issue recently.

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
```diff
@@ -547,16 +548,17 @@ class VariantContextToInternalRowConverter(
...
-          } else {
-            provideWarning(
-              s"Key $key has empty value $valueStr, but FLAG is not supported in FORMAT fields."
-            )
+          val value = genotype.getExtendedAttribute(key)
+          value match {
+            case _: JBoolean =>
+              provideWarning(
+                s"Key $key has an empty value, but FLAG is not supported in FORMAT fields."
+              )
```
kianfar77: Why is this warning associated with JBoolean type?

core/src/main/scala/io/projectglow/vcf/VariantContextToVCFRowConverter.scala
```diff
@@ -16,46 +16,38 @@
...
-import htsjdk.variant.vcf.{VCFConstants, VCFHeader}
+import htsjdk.variant.vcf.{VCFEncoderUtils, VCFHeader}
 
 import io.projectglow.common.{GlowLogging, VCFRow}
 
-// HTSJDK VariantContext -> VCFRow
-// Based on the HTSJDK classes VCFEncoder and CommonInfo
 object VariantContextToVCFRowConverter {
 
+  // Usually encodes the object normally, but does not pretty-print doubles
```
kianfar77: My concern here is if we read a VCF and write it back again we should get exactly the same file (perhaps excluding the header). Does this change make sure that happens? It would be problematic for the used if for example they read a file containing doubles and write them back to disk and they get some other formatting of doubles. Your added test on parsedObjectAsString shows that things work correctly for that function. 

But I do not remember if our unit tests verify 100% conformance of read VCF and written one, e.g., whether their assert statements compare the verbatim strings from written file with the original file or they themselves do some formatting and hence mess up the comparison. Perhaps it is a good idea to read and write back an example VCF and compare visually to make sure things work correctly as this was not the case before https://github.com/projectglow/glow/commit/15e102e984a337fbd13a023d7c0e71dc582b7dcc

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
karenfeng: FORMAT fields should not have a FLAG type (boolean value); I'll change the message to make it clearer.

core/src/main/scala/io/projectglow/vcf/VariantContextToVCFRowConverter.scala
karenfeng: These are tested via round-trips in the `VCFFileWriterSuite`s. In particular, the `Read single sample VCF with VCF parser` test compares the value of the `Double` field `ReadPosRankSum=1.075` (the usual VCF encoder prettifies this as `1.08`). `Read VEP VCF with VCF parser` compares the value of an `Array(Double)` field as well.

kianfar77:
LGTM

PR #125, opened 2020-01-07
Unflake VCFPiperSuite

henrydavidge:

I believe that the test could previously fail because the output formatter would throw an exception while the input formatter still had a stream open. Now, we wait for all threads to be shut down before we do the leaked stream check.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


core/src/test/scala/io/projectglow/vcf/VCFPiperSuite.scala
```diff
@@ -268,7 +255,7 @@ class VCFPiperSuite extends GlowBaseTest {
     assertThrows[IllegalArgumentException](Glow.transform("pipe", inputDf, options))
   }
 
-  test("output validation stringency") {
+  test(s"output validation stringency") {
```
karenfeng: Huh

PR #128, opened 2020-01-08
Record one big VCF write event per file

karenfeng:

Ensures that we only log one big VCF write per file, not per partition. For ease of use down the road, also simplifies the logging interface.

## How is this patch tested?
- [ ] Unit tests
- [x] Integration tests
- [ ] Manual tests

Events are propagated on the universe side.

### DIFF
--- a/core/src/main/scala/io/projectglow/bgen/BigBgenDatasource.scala
+++ b/core/src/main/scala/io/projectglow/bgen/BigBgenDatasource.scala
@@ -52,7 +52,7 @@ object BigBgenDatasource extends HlsEventRecorder {
     BigBgenOptions(bitsPerProb, maxPloidy, defaultPloidy, defaultPhasing)
   }
 
-  private def logWrite(parsedOptions: BigBgenOptions): Unit = {
+  private def logBgenWrite(parsedOptions: BigBgenOptions): Unit = {
     val logOptions = Map(
       BITS_PER_PROB_KEY -> parsedOptions.bitsPerProb,
       MAX_PLOIDY_KEY -> parsedOptions.maxPloidy,
@@ -67,7 +67,7 @@ object BigBgenDatasource extends HlsEventRecorder {
     val numVariants = data.count
     val parsedOptions = parseOptions(options)
 
-    logWrite(parsedOptions)
+    logBgenWrite(parsedOptions)
 
     data.queryExecution.toRdd.mapPartitionsWithIndex {
       case (idx, it) =>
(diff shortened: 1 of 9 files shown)


core/src/main/scala/io/projectglow/bgen/BigBgenDatasource.scala
```diff
@@ -22,48 +22,52 @@ import org.apache.spark.rdd.RDD
...
+  def parseOptions(options: Map[String, String]): BigBgenOptions = {
     val bitsPerProb = options.getOrElse(BITS_PER_PROB_KEY, BITS_PER_PROB_DEFAULT_VALUE).toInt
     val maxPloidy = options.getOrElse(MAX_PLOIDY_KEY, MAX_PLOIDY_VALUE).toInt
     val defaultPloidy = options.getOrElse(DEFAULT_PLOIDY_KEY, DEFAULT_PLOIDY_VALUE).toInt
     val defaultPhasing = options.getOrElse(DEFAULT_PHASING_KEY, DEFAULT_PHASING_VALUE).toBoolean
+    BigBgenOptions(bitsPerProb, maxPloidy, defaultPloidy, defaultPhasing)
+  }
 
-    // record bgenWrite event in the log
+  def logWrite(parsedOptions: BigBgenOptions): Unit = {
```
kianfar77: nit: logBgenWrite is better in terms of naming consistency.

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
```diff
@@ -94,21 +91,14 @@ class NormalizeVariantsTransformer extends DataFrameTransformer with HlsUsageLog
...
-  private[projectglow] def logNormalizeVariants(mode: String): Unit = {
-    val logOptions = Map(MODE_KEY -> mode)
-    recordHlsUsage(
-      HlsMetricDefinitions.EVENT_HLS_USAGE,
-      Map(
-        HlsTagDefinitions.TAG_EVENT_TYPE -> HlsTagValues.EVENT_NORMALIZE_VARIANTS
-      ),
-      blob = hlsJsonBuilder(logOptions)
-    )
+  def logNormalizeVariants(mode: String): Unit = {
```
kianfar77: nit: Now that we have recordHlsEvent, I think it cleaner if we drop this function and plug recordHlsEvent with appropriate arguments in the the three cases above directly.

PR #134, opened 2020-01-17
Implement splitter as a separate transformer

kianfar77:
This PR separates split_multiallelics transformer from normalize_variants transformer and deprecates the "mode" option which allowed running splitter as an option in the normalize_variants transformer.  

## How is this patch tested?
- [X] Unit tests
- [ ] Integration tests
- [X] Manual tests

(Details)


core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
```diff
@@ -16,73 +16,90 @@
...
+      VariantNormalizer.normalize(
+        df,
+        options.get(REFERENCE_GENOME_PATH)
+      )
+    }
+  }
 
-    val validationStringency: ValidationStringency = VCFOptionParser
-      .getValidationStringency(options)
+  /** The following function is for backward compatibility to the previous API where
```
henrydavidge: nit: text should start on second line
```
/**
 * The following...
```

Issue #117, opened 2019-12-23
Support Scala 2.12


hammer:
Spark has had [2.12 support](https://issues.apache.org/jira/browse/SPARK-14220) for over a year; what's blocking Glow from also supporting 2.12?

henrydavidge:
Nothing blocking it, just don't have it yet. We'll certainly add 2.12 in advance of Spark 3.0 and would happily take a contribution sooner.

henrydavidge:
There's a 2.12 artifact in the 0.3.0 release. Closing.

PR #137, opened 2020-01-21
Cross build for Scala 2.11 and 2.12

karenfeng:

Addresses https://github.com/projectglow/glow/issues/117.

Implements support for Scala 2.12.8 (on top of 2.11.12).

There is some messiness associated with the fact that `scalalogging` refactored their logging code out of `slf4j` between 2.11 and 2.12.

## How is this patch tested?
- [ ] Unit tests
- [x] Integration tests
- [x] Manual tests

To run all of the Scala, Python, and docs tests (no staged release tests) on Scala 2.11 and Scala 2.12, run `sbt test`.

All of the tests can be run against Scala 2.11 and Scala 2.12.

|    | Scala 2.11 | Scala 2.12 | 
| --- | --- | --- |
| **Scala tests** | core_2_11 | core_2_12 |
| **Python tests** | python_2_11 | python_2_12 |
| **Docs tests** | docs_2_11 | docs _2_12 |
| **Staged release tests** | stagedRelease_2_11 | stagedRelease_2_12 |

To release both builds: `sbt release`.


.circleci/config.yml
```diff
@@ -32,7 +32,11 @@ jobs:
           environment:
           command: |
             export PATH=$HOME/conda/envs/glow/bin:$PATH
-            sbt printTests coverage "testOnly $(circleci tests split --split-by=timings --timings-type=classname core-test-names.log | tr '\n' ' ') --" coverageReport exit
+            if [[ $CIRCLE_NODE_INDEX == 0 ]]; then
```
henrydavidge: What's up with this?

build.sbt
```diff
@@ -3,11 +3,15 @@ import scala.sys.process._
...
 
-ThisBuild / scalaVersion := s"$scalaMajorMinor.12"
+lazy val scala212 = "2.12.8"
+lazy val scala211 = "2.11.12"
+lazy val supportedScalaVersions = List(scala212, scala211)
+
+ThisBuild / scalaVersion := scala211
 ThisBuild / organization := "io.projectglow"
 ThisBuild / scalastyleConfig := baseDirectory.value / "scalastyle-config.xml"
+ThisBuild / crossScalaVersions := Nil
```
henrydavidge: Hm, I would expect this to be scala211 and scala212. Could you please comment why it's like this?

build.sbt
```diff
@@ -167,6 +174,7 @@ lazy val pythonSettings = Seq(
   publish / skip := true
 )
 
+// Python tests only work on Scala 2.11
```
henrydavidge: Do they only work on 2.11? Or do we only run them on 2.11?

build.sbt
```diff
@@ -257,18 +267,22 @@ lazy val stagedRelease = (project in file("core/src/test")).settings(
 
 import ReleaseTransformations._
 
+// Don't use sbt-release's cross facility
```
henrydavidge: Why?

.circleci/config.yml
karenfeng: I'm running the 2.11 tests on the first node, and 2.12 on the second.

henrydavidge: Is it common to run tests against both Scala versions on every PR? What does Spark do?

karenfeng: For each PR on 2.4, it looks like they run the tests on the matrix of Scala 2.11/2.12 and Hadoop 2.6/2.7.

PR #140, opened 2020-01-22
Simplified linear regression

henrydavidge:

This PR changes our linear regression implementation to use an algorithm similar to Hail's. In addition to being much more concise, this implementation performs better with a large number of covariates because 1) it avoids copying the covariate matrix for each new row 2) it does not require inverting the full X matrix to compute the standard error.

On 5M samples and 1M variants, this change improves the performance of linear regression on a 16 worker i3.4xlarge cluster from ~4h to ~2-2.25h.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

(Details)


core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
```diff
@@ -35,9 +35,9 @@ import io.projectglow.sql.util.ExpectsGenotypeFields
  */
 object VariantUtilExprs {
   def genotypeStates(genotypes: ArrayData, genotypesSize: Int, genotypesIdx: Int): ArrayData = {
-    val output = new Array[Int](genotypes.numElements())
+    val output = ArrayData.allocateArrayData(IntegerType.defaultSize, genotypes.numElements(), "")
```
karenfeng: OOC, did this result in a speedup?

core/src/test/scala/io/projectglow/tertiary/LinearRegressionSuite.scala
```diff
@@ -138,13 +148,11 @@ class LinearRegressionSuite extends GlowBaseTest {
...
     val apacheResults = timeIt("Apache linreg") {
-      testData.genotypes.map { g =>
-        olsBaseline(g.toArray, testData.phenotypes.toArray, testData.covariates.toArray)
-      }
+      testDataToOlsBaseline(testData)
     }
 
     val ourResults = timeIt("DB linreg") {
-      if (useSpark) {
+      if (true) {
```
karenfeng: I assume this was for testing only?

core/src/test/scala/io/projectglow/tertiary/LinearRegressionSuite.scala
```diff
@@ -155,22 +163,25 @@ class LinearRegressionSuite extends GlowBaseTest {
         spark
           .createDataFrame(rows)
           .withColumn("id", monotonically_increasing_id())
-          .repartition(10)
+          .repartition(1)
```
karenfeng: Why?

core/src/test/scala/io/projectglow/tertiary/LinearRegressionSuite.scala
```diff
@@ -304,6 +320,24 @@ class LinearRegressionSuite extends GlowBaseTest {
     assert(e.getCause.getMessage.contains(error))
   }
 
+  // Our linear regression algorithm projects the genotypes onto orthogonal complement of the
+  // covariate vector space. When the genotypes are a linear combination of some covariates, this
+  // projection is 0, so beta, standard error, and p value are all undefined. However, finite
+  // precision can instead cause the error to be massive with respect to beta.
+  test("negligible p value if genotypes are in covariate span") {
```
karenfeng: Nit: this is a misleading name; negligible suggests the p-value is small when it's actually very large.

core/src/test/scala/io/projectglow/tertiary/LinearRegressionSuite.scala
```diff
@@ -304,6 +320,24 @@ class LinearRegressionSuite extends GlowBaseTest {
...
+    val genotypes = twoDArrayToBreezeMatrix(testData.covariates.toArray)(::, 1)
+    val phenotypes = new DenseVector[Double](testData.phenotypes.toArray)
+    val ctx = CovariateQRContext.computeQR(twoDArrayToSparkMatrix(testData.covariates.toArray))
+
+    assertThrows[SingularMatrixException] {
+      olsBaseline(genotypes.toArray, phenotypes.toArray, testData.covariates.toArray)
+    }
+
+    val results = LinearRegressionGwas.runRegression(genotypes, phenotypes, ctx)
+    assert(results.pValue > 0.999999 || results.pValue.isNaN)
```
karenfeng: Does the p-value change?

core/src/main/scala/io/projectglow/sql/expressions/LinearRegressionGwas.scala
```diff
@@ -32,173 +30,48 @@ object LinearRegressionGwas extends GlowLogging {
...
-      phenotypes,
-      covariateQRContext.numCols,
-      1
-    )
-
-    // Calculate the t statistic described in
-    // https://web.stanford.edu/~hastie/ElemStatLearn/printings/ESLII_print12.pdf equation
-    // 3.12, we need to invert (G^T * G)
-    // Note that (G^T * G) = (R_G^T * R_G)
-    val qrCopy = covariateQRContext.values.clone()
```
karenfeng: To clarify, were these two lines the expensive ones?

core/src/test/scala/io/projectglow/tertiary/LinearRegressionSuite.scala
henrydavidge: Yes, it can be NaN.

core/src/test/scala/io/projectglow/tertiary/LinearRegressionSuite.scala
henrydavidge: Oops!

core/src/test/scala/io/projectglow/tertiary/LinearRegressionSuite.scala
henrydavidge: Oops!

core/src/main/scala/io/projectglow/sql/expressions/LinearRegressionGwas.scala
henrydavidge: I think so. I couldn't actually see this is in the profile directly, but i believe that this led to more GC and worse cache locality.

core/src/test/scala/io/projectglow/tertiary/LinearRegressionSuite.scala
henrydavidge: I believe that NaN is actually the proper value since `qtx` should be 0, so all computations include a `0/0` term.

core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
henrydavidge: yes, slightly.

PR #143, opened 2020-01-24
Automatically test staged releases

karenfeng:

During a release, we now automatically run our unit tests against the staged Maven artifact with minimal imports. This will be used to ensure that the released artifact functions as anticipated before we promote it to Central.

## How is this patch tested?
- [x] Unit tests
- [x] Integration tests
- [ ] Manual tests

`stagedRelease/test`


build.sbt
```diff
@@ -235,19 +241,33 @@ ThisBuild / publishMavenStyle := true
 ThisBuild / bintrayOrganization := Some("projectglow")
 ThisBuild / bintrayRepository := "glow"
 
+val stableVersion = settingKey[String]("Stable version")
+ThisBuild / stableVersion := IO.read((ThisBuild / baseDirectory).value / "stable-version.txt").trim()
+
+lazy val stagedRelease = (project in file("core")).settings(
```
henrydavidge: Does this not mean that we're using the managed source directory from `core`? I don't think we want that, do we?

karenfeng: Not quite - by setting `unmanagedSourceDirectories in Compile := Nil`, we are not using any code for compilation.

karenfeng: To clarify, `unmanagedSourceDirectories` consists of `scalaSource` and `javaSource`: https://www.scala-sbt.org/1.x/docs/Howto-Customizing-Paths.html#Add+an+additional+source+directory

henrydavidge: Ah, got it. Why do we copy the settings from `core` then? It might be more clear to just base this on `commonSettings`

.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           environment:
           command: |
             export PATH=$HOME/conda/envs/glow/bin:$PATH
-            sbt printTests coverage "testOnly $(circleci tests split --split-by=timings --timings-type=classname core-test-names.log | tr '\n' ' ') --" coverageReport exit
+            sbt printTests coverage "core/testOnly $(circleci tests split --split-by=timings --timings-type=classname core-test-names.log | tr '\n' ' ') --" coverageReport exit
```
henrydavidge: This makes sense to me, but it is this change necessary for this PR? I'm fine with adding it now, just trying to understand.

build.sbt
```diff
@@ -235,6 +244,17 @@ ThisBuild / publishMavenStyle := true
...
 
+val stableVersion = settingKey[String]("Stable version")
+ThisBuild / stableVersion := IO.read((ThisBuild / baseDirectory).value / "stable-version.txt").trim()
+
+lazy val stagedRelease = (project in file("core/src/test")).settings(
+  commonSettings,
+  resourceDirectory in Test := baseDirectory.value / "resources",
+  scalaSource in Test := baseDirectory.value / "scala",
+  libraryDependencies ++= providedDependencies ++ testDependencies :+ "io.projectglow" %% "glow" % stableVersion.value,
+  resolvers += "bintray-staging" at "https://dl.bintray.com/projectglow/glow"
```
henrydavidge: What is the priority here? I think we'd want to make sure that the bintray repo has priority over a possible glow artifact that could exist upstream. (Although I guess it is an error anyway to stage an already released version).

karenfeng: Good call, I'll just set it to have a single resolver.

.circleci/config.yml
karenfeng: Ah whoops, I forgot to change this back after I set `root`.

Issue #109, opened 2019-12-13
Memory utilization of variant normalization


emaxwell:
The current implementation of normalize_variants is rather memory intensive, presumably because of the interfacing with HTSJDK. Optimizing this function to use more pure Spark machinery could yield a nice performance improvement.

kianfar77:
Resolved by #146. About 2.5x speed up was observed on some tests.

Issue #108, opened 2019-12-13
Glow transform functions drop supplemental columns.


emaxwell:
Currently, transform functions like 'normalize_variants' drop data in columns that are not part of the core VCF/BGEN schema by setting the values to null. Only columns of strings having the "INFO_" prefix are maintained if they are not part of the core schema. It would be ideal if Glow transformers could maintain these supplemental columns from the input DataFrame.

kianfar77:
Resolved by #146

PR #146, opened 2020-01-29
[HLS-407] Implement Variant Normalizer as a Spark Sql expression instead of using VariantContext

kianfar77:

This PR changes the implementation of  normalize_variants transformer in order to avoid use of VariantContext. Instead it uses a spark sql expression called normalize_variant to perform variant normalization transformation.

The reference genome now uses htsjdk.samtools.reference.IndexedFastaSequenceFile instead of gatk's org.broadinstitute.hellbender.engine.ReferenceDataSource.

In addition, a normalizationFlag column in now added to the dataframe after normalization to indicate the status of normalization attempt.

The replace_columns option is also added to the normalize_variants transformer to give the user control on whether to add or replace the normalized start, end, refAllele, and altAllele columns. 

Addresses #108 and #109 

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

(Details)


core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
```diff
@@ -0,0 +1,99 @@
...
+      altAlleles.asInstanceOf[Array[String]],
+      state.get()
+    )
+  }
+}
+
+
+case class NormalizeVariant(contigName: Expression,
+                            start: Expression,
+                            end: Expression,
```
henrydavidge: Do we actually use `end` here?

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
```diff
@@ -0,0 +1,99 @@
...
+                            altAlleles: Expression,
+                            refGenomePathString: Expression)
+  extends SenaryExpression
+    with ImplicitCastInputTypes {
+
+  override def prettyName: String = "normalize_variant"
+
+  override def dataType: DataType = VariantNormalizer.normalizationSchema
+
+  override def inputTypes: Seq[DataType] =
```
henrydavidge: I think `start` and `end` are usually longs?

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
```diff
@@ -0,0 +1,99 @@
...
+
+    if (state.get() == null) {
+      // Save fasta sequence file
+      val refGenomeIndexedFasta = new IndexedFastaSequenceFile(Paths.get(refGenomePathString.asInstanceOf[String]))
+      state.set(refGenomeIndexedFasta)
+      TaskContext.get().addTaskCompletionListener(_ => state.remove())
+    }
+
+    VariantNormalizer.normalizeVariant(
+      contigName.asInstanceOf[String],
```
henrydavidge: I think these are going to be `UTF8String`s

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
kianfar77: Based off of end to avoid interfering with the user's choice in definition of end.

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
kianfar77: WIP effect. fixed.

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
kianfar77: WIP effect. Fixed.

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
```diff
@@ -0,0 +1,112 @@
...
+      refAllele: Any,
+      altAlleles: Any,
+      refGenomePathString: Any): InternalRow = {
+
+    if (state.get() == null) {
+      // Save fasta sequence file
+      val refGenomeIndexedFasta = new IndexedFastaSequenceFile(
+        Paths.get(refGenomePathString.asInstanceOf[UTF8String].toString))
+      state.set(refGenomeIndexedFasta)
+      TaskContext.get().addTaskCompletionListener[Unit](_ => state.remove())
```
henrydavidge: Do you not need to close the reference file?

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
```diff
@@ -16,318 +16,182 @@
...
 
   }
 
   private val PAD_WINDOW_SIZE = 100
 
+  val FLAG_ERROR = "Error"
+  val FLAG_CHANGED = "Changed"
+  val FLAG_UNCHANGED = "Unchanged"
+
+  val normalizedStartField = StructField("normalizedStart", LongType)
```
henrydavidge: Hm, I don't love adding the `normalized` prefix to all of these column names. Since this expression returns a struct, it's simple to keep the fields grouped together. We can use the normal names (`start`, `end`, etc). In the case where we don't replace columns, we can just call this struct `normalizationResult` or something like that.

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
```diff
@@ -16,318 +16,182 @@
...
+      UTF8String.fromString(allAlleles(0)),
+      ArrayData.toArrayData(allAlleles.tail.map(UTF8String.fromString(_))),
+      UTF8String.fromString(flag)
     )
 
   }
 
   private val PAD_WINDOW_SIZE = 100
 
+  val FLAG_ERROR = "Error"
```
henrydavidge: We should be consistent about how we report errors. In liftover, I think that we return an error message in the case of failure. If the error message is null, the operation was successful.

henrydavidge: Whether the variant was changed or not should be communicated with a boolean

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
```diff
@@ -16,318 +16,182 @@
...
-    while (alleles.forall(
-        a =>
-          a.getBases()(a.length() - nTrimmedBasesBeforeNextPadding - 1) ==
-          firstAlleleBaseFromRight
-      )) {
-      // Last base in all alleles are the same
+    if (refAllele.isEmpty) {
+      // if no alleles, throw exception
+      logger.info("No REF and ALT alleles...")
+      flag = FLAG_ERROR
```
henrydavidge: Is this the only error condition? What else can go wrong during normalization?

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
henrydavidge: Also, what's in the rest of the result fields when normalization fails?

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
kianfar77: Forgot. Added.

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
```diff
@@ -16,318 +16,182 @@
...
+    } else if (refAllele.length == 1 && altAlleles.forall(_.length == 1)) {
+      // if a SNP, do nothing
+      flag = FLAG_UNCHANGED
+    } else if (altAlleles.exists(_.matches(".*[<|>|*].*"))) {
```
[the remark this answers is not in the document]
kianfar77: < and > are standard. I have seen * used as well. Do you have any other suggestion?

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
kianfar77: Changed to what you mention.

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
kianfar77: Changed to what you mention. Have the results to be null in case of error.

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
```diff
@@ -0,0 +1,119 @@
...
+  }
+}
+
+case class NormalizeVariantExpr(
+    contigName: Expression,
+    start: Expression,
+    end: Expression,
+    refAllele: Expression,
+    altAlleles: Expression,
+    refGenomePathString: Expression)
```
henrydavidge: We should check that this expression if foldable

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
```diff
@@ -17,72 +17,122 @@
...
+
+      val modeOption = options.get(MODE_KEY).map(StringUtils.toSnakeCase)
+      backwardCompatibleTransform(df, refGenomePathString, replaceColumns, modeOption)
+
     } else {
+
       recordHlsEvent(HlsTagValues.EVENT_NORMALIZE_VARIANTS)
 
-      VariantNormalizer.normalize(
+      if (refGenomePathString.isEmpty) {
```
henrydavidge: Check this in the expression

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
```diff
@@ -17,72 +17,122 @@
...
-import io.projectglow.transformers.splitmultiallelics.VariantSplitter
+import org.apache.spark.sql.functions._
 
 /**
- * Implements DataFrameTransformer to transform the input DataFrame of varaints to an output
- * DataFrame of normalized variants (normalization is as defined in bcftools norm).
+ * Implements DataFrameTransformer to transform the input DataFrame of variants to an output
+ * DataFrame of normalized variants (normalization is as defined in vt normalize or bcftools norm).
+ *
+ * A path to reference genome containing .fasta, and .fai files must be provided
```
henrydavidge: Is this path to the fasta file or do a directory containing both?

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
```diff
@@ -17,72 +17,122 @@
...
+ *       this field will be null.
+ *
+ * If the replace_columns option is false, the transformer also adds the StructType
+ * normalizationResult column to the DataFrame which contains the normalized start, end,
+ * referenceAllele, and alternateAlleles columns, whether changed or unchanged.  In case of
+ * error, normalizationResult will be null.
+ *
+ * If the replace_columns option is true (default), the transformer replaces the original start,
+ * end, referenceAllele, and alternateAlleles columns with the fields of normalizationResult (for
+ * no-error rows) and drops the normalizationResult column.
```
henrydavidge: Hm, it seems pretty dangerous to just drop all of the error rows. AFAIK, most other tools right out errors separately. The Spark way of doing that is to leave the errors so that people can filter out the errors / successes.

functions.yml
```diff
@@ -116,7 +116,25 @@ etl:
           doc: Minimum fraction of bases that must remap to lift over successfully
           type: double
           is_optional: true
-          
+    - name: normalize_variant
+      doc: Normalizes the variant like vt decompose -s
```
henrydavidge: Can you improve the documentation? It would be nice to have a description of the algorithm as well as details about what the different result fields look like if there is / is not an error. It's actually better to put documentation here than on the scala transformer class.

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
kianfar77: Done

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
kianfar77: Clarified.

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
kianfar77: No row will be dropped. The erroneous rows will not be touched and the normalizationStatus row is there to detect that they had errors.

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
kianfar77: Moved

functions.yml
kianfar77: Added more doc here. Kept the one in transformer as well because it has that replace_columns option which makes the doc a bit different.

kianfar77:
> Getting closer! I still have some questions on the API and user experience.

Please see my responses.

functions.yml
```diff
@@ -116,7 +116,36 @@ etl:
           doc: Minimum fraction of bases that must remap to lift over successfully
           type: double
           is_optional: true
-          
+    - name: normalize_variant
+      doc: "Normalize the variant using algorithms similar to vt normalize or bcftools norm. Adds
```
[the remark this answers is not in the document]
henrydavidge: This renders a little strangely in the python / scala files. I think you you want to use a multiline YAML string:
```
doc: |
blah
blah
blah
```

And then you might have to make a change to the templates like this one: https://github.com/henrydavidge/glow/pull/new/pydoc

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
```diff
@@ -16,318 +16,212 @@
...
   }
 
   private val PAD_WINDOW_SIZE = 100
 
+  val normalizationResultFieldName = "normalizationResult"
+  val normalizationStatusFieldName = "normalizationStatus"
+  val changedFieldName = "changed"
+  val errorMessageFieldName = "errorMessage"
+
+  val normalizationStatusStructField = StructField(
```
henrydavidge: Hm, what is the upside of nesting everything in `normalizationResult` / `normalizationStatus` vs a flat struct? I generally think it's more user friendly to avoid nesting columns since it makes fields easier to access.

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
```diff
@@ -93,6 +145,65 @@ class NormalizeVariantsTransformer extends DataFrameTransformer with HlsEventRec
...
+        )
+        .drop(normalizationResultFieldName)
+
+    } else {
+      dfNormalized
+    }
+  }
+
+  val REFERENCE_GENOME_PATH = "reference_genome_path"
+  val REPLACE_COLUMNS = "replace_original_columns"
```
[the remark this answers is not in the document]
henrydavidge: Ah, it looks like the docs already refer to this field as `replace_columns` as well.

core/src/main/scala/io/projectglow/sql/expressions/NormalizeVariantExpr.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  }
+
+  override def doGenCode(ctx: CodegenContext, ev: ExprCode): ExprCode = {
+    nullSafeCodeGen(
+      ctx,
+      ev,
+      (contigName, start, end, refAllele, altAlleles, refGenomePathString) => {
+        s"""
+           |
```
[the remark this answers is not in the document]
kianfar77: fixed

core/src/main/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformer.scala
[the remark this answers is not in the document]
kianfar77: yes. typo.

core/src/main/scala/io/projectglow/transformers/normalizevariants/VariantNormalizer.scala
kianfar77: I put the start, end, ref, and alts into normalizationResult as you suggested. normalizationStatus is another column. So there is only one level of nesting. 

I felt result separate from status makes more sense as it makes moving the actual result data columns, i.e., normalized (start, end, ref, alts), altogether easier after normalization. Same goes for dropping the status column after verifying no error happened or which variants changed.

functions.yml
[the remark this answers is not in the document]
kianfar77: Done. Except for the * at the beginning of each line in scala functions.

PR #151, opened 2020-02-11
Python documentation and tests

henrydavidge:
- Adds a new `examples` dictionary field to each function definition in `functions.yml`. Currently, we only use the `python` example from this dictionary. The python example is added to the docstring for the generated function. An example is mandatory for all new functions that are included in the python API.
- Sets up doctests for the python package, both for the generated functions and for the top level functions.
- Adds python documentation using the sphinx `autodoc` feature
- A number of build tweaks to make it easier to work with the generated clients. The input files (`functions.yml`, the template file, and the generator script) are now properly tracked by sbt, so we always regenerate the output when any of them change. There's now a `generateFunctions` sbt task that quickly generates both the scala and python files.

I did NOT add better documentation (except for the example) for the functions that are already included. We'll need to do that in a followup.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

You can run `make livehtml` to check it out!

Screen Shot 2020-02-12 at 11 32 31 AM

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -204,7 +204,7 @@ lazy val python =
       test in Test := {
         // Pass the test classpath to pyspark so that we run the same bits as the Scala tests
         val ret = Process(
-          Seq("pytest", "python"),
+          Seq("pytest", "--doctest-modules", "python"),
           None,
           "SPARK_CLASSPATH" -> sparkClasspath.value,
           "SPARK_HOME" -> sparkHome.value
--- a/build.sbt
+++ b/build.sbt
@@ -208,7 +208,7 @@ lazy val pythonSettings = Seq(
   libraryDependencies ++= testSparkDependencies,
   sparkClasspath := (fullClasspath in Test).value.files.map(_.getCanonicalPath).mkString(":"),
   sparkHome := (ThisBuild / baseDirectory).value.absolutePath,
-  pythonPath := ((ThisBuild / baseDirectory).value / "python" / "glow").absolutePath,
+  pythonPath := ((ThisBuild / baseDirectory).value / "python").absolutePath,
   publish / skip := true
 )
 
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,8 @@
 import scala.sys.process._
 
 import sbt.Tests._
+import sbt.Keys._
+import sbt.nio.Keys._
 
 val sparkVersion = "2.4.3"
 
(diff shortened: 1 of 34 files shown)


build.sbt
```diff
@@ -246,7 +253,7 @@ lazy val docs = (project in file("docs"))
     }
   )
   .cross
-  .dependsOn(core % "test->test")
+  .dependsOn(core % "test->test", python)
```
karenfeng: Why are we depending on `python`?

build.sbt
```diff
@@ -199,30 +208,28 @@ lazy val pythonSettings = Seq(
...
 lazy val python =
   (project in file("python"))
     .settings(
       pythonSettings,
+      functionGenerationSettings,
       test in Test := {
         // Pass the test classpath to pyspark so that we run the same bits as the Scala tests
         val ret = Process(
-          Seq("pytest", "python"),
+          Seq("pytest", "--doctest-modules", "python"),
```
karenfeng: Cool, so this tests docstrings?

core/src/test/resources/log4j.properties
```diff
@@ -1,5 +1,5 @@
 # Set everything to be logged to the console
-log4j.rootCategory=INFO, file
+log4j.rootCategory=TRACE, file
```
karenfeng: What are these changes for?

docs/source/api-docs/index.rst
```diff
@@ -0,0 +1,31 @@
...
+  glow.transform('pipe', df, cmd='["cat"]', input_formatter='vcf', output_formatter='vcf',
+      in_vcf_header='infer')
+  df.select(glow.genotype_states('genotypes'))
+
+Modules
+~~~~~~~
+
+.. toctree::
+
+  functions
```
karenfeng: This doesn't point to anything right now - do you mean something along the lines of the following, or would that be a followup PR?
```
.. automodule:: glow.functions
  :members:
```

python/glow/functions.py.TEMPLATE
```diff
@@ -17,13 +17,15 @@ def {{ function.name }}({{ function.args|map('fmt_python_signature')|join(', ')
     """
     {{ function.doc }}
 
-    Parameters
-    ----------
+    Added in version {{ function.since }}.
```
karenfeng: Why did we change these?

python/render_template.py
```diff
@@ -60,10 +60,11 @@ def fmt_python_signature(value):
 
 def fmt_python_call(value):
     if value.get('is_var_args'):
-        return f'_to_seq({value["name"]})'
+        converter = ', _to_java_column' if not 'type' in value else ''
```
karenfeng: Can you explain these changes?

henrydavidge: Done

Issue #176, opened 2020-03-23
Support for block-gzipped fasta files


Hoeze:
Trying to run the "normalize_variants" transformer with a block-gzipped + indexed fasta file results in the following error:
```java
Py4JJavaError: An error occurred while calling o376.collectToPython.
: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 5.0 failed 1 times, most recent failure: Lost task 0.0 in stage 5.0 (TID 35, localhost, executor driver): htsjdk.samtools.SAMException: Indexed block-compressed FASTA file cannot be handled: /s/project/variantDatabase/gtex/v8/hg19/dna.fa.gz
	at htsjdk.samtools.reference.IndexedFastaSequenceFile.<init>(IndexedFastaSequenceFile.java:84)
	at htsjdk.samtools.reference.IndexedFastaSequenceFile.<init>(IndexedFastaSequenceFile.java:98)
	at io.projectglow.sql.expressions.NormalizeVariantExpr$.doVariantNormalization(NormalizeVariantExpr.scala:47)
	at io.projectglow.sql.expressions.NormalizeVariantExpr.doVariantNormalization(NormalizeVariantExpr.scala)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anonfun$13$$anon$1.hasNext(WholeStageCodegenExec.scala:636)
	at org.apache.spark.sql.execution.SparkPlan$$anonfun$2.apply(SparkPlan.scala:255)
	at org.apache.spark.sql.execution.SparkPlan$$anonfun$2.apply(SparkPlan.scala:247)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitionsInternal$1$$anonfun$apply$24.apply(RDD.scala:836)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitionsInternal$1$$anonfun$apply$24.apply(RDD.scala:836)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:324)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:288)
[...]

PR #177, opened 2020-03-26
Support indexed bgzipped fasta in variant normalizer

kianfar77:

Due to use of ``htsjdk.samtools.reference.IndexedFastaSequenceFile`` to provide the reference genome to variant normalizer, blocked-gzipped-indexed files were not supported. This is now replaced with ``htsjdk.samtools.reference.ReferenceSequenceFileFactory`` which can handle both uncompressed fasta (with .fai) and block-gzipped fasta (with .fai and .gzi). 

Unit tests were added to check normalization works with both uncompressed and bgzipped fastas.

Addresses #176 

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


kianfar77:
Added a couple of more tests for exceptions at nonexistent path and reference file without index.

core/src/test/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformerSuite.scala
```diff
@@ -199,12 +204,50 @@ class NormalizeVariantsTransformerSuite extends GlowBaseTest with GlowLogging {
...
+    assertThrows[SparkException](
+      testNormalizedvsExpected(
+        vtTestVcfMultiAllelic,
+        vtTestVcfMultiAllelicExpectedNormalized,
+        Option(vtTestReferenceNonexistent)
+      )
+    )
+  }
+
+  gridTest("Reference not indexed ")(
```
karenfeng: Nit: trailing space

core/src/test/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformerSuite.scala
```diff
@@ -199,12 +204,50 @@ class NormalizeVariantsTransformerSuite extends GlowBaseTest with GlowLogging {
...
+      testNormalizedvsExpected(
+        vtTestVcfMultiAllelic,
+        vtTestVcfMultiAllelic,
+        None
+      )
+    )
+  }
+
+  test("Invalid reference path ") {
+    assertThrows[SparkException](
```
karenfeng: For these tests, can we check the exception's error message to make sure the user sees something reasonable?

kianfar77: I have checked. They see proper messages from htsjdk.

karenfeng: Could we catch the error messages with `intercept` to verify in the unit tests?

core/src/test/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformerSuite.scala
kianfar77: Fixed.

core/src/test/scala/io/projectglow/transformers/normalizevariants/NormalizeVariantsTransformerSuite.scala
kianfar77: Done

Issue #172, opened 2020-03-13
Support for Spark 3.0


raghavmi:
Since Spark 3.0 is right around the corner, can you folks comment on your plans to support it?

karenfeng:
We will release Glow for Spark 3.0 when the new major version is released, and have prepared the cross-build on the nightly snapshots.

raghavmi:
That’s great!

Thanks for the update.

PR #181, opened 2020-03-31
Run nightly tests against Spark 3

karenfeng:

Simplifies the workflow to build/test on different versions on Spark, so the developer only needs to modify the environment variable `SPARK_VERSION`.

Also creates a per-job and nightly trigger to make sure we're compatible with the nightly Spark 3 snapshot release, and the most recent PySpark release (dev2).

Related to https://github.com/projectglow/glow/issues/172.

## How is this patch tested?
- [x] Unit tests
- [x] Integration tests
- [ ] Manual tests


.circleci/config.yml
```diff
@@ -131,10 +131,45 @@ jobs:
...
+          keys:
+            - conda-deps-v1-{{ checksum "python/environment.yml" }}
+      - *install_deps
+      - run:
+          name: Change Spark version
+          environment:
+          command: |
+            echo "3.0.0-SNAPSHOT" > spark-version.txt
+      - run:
+          name: Run Scala tests
```
henrydavidge, line 148: Seems like we should bump the pyspark version and run Python tests as well

.circleci/config.yml
```diff
@@ -131,10 +131,45 @@ jobs:
...
+      - checkout
+      - restore_cache:
+          keys:
+            - conda-deps-v1-{{ checksum "python/environment.yml" }}
+      - *install_deps
+      - run:
+          name: Change Spark version
+          environment:
+          command: |
+            echo "3.0.0-SNAPSHOT" > spark-version.txt
```
henrydavidge, line 146: Why read this from a text file? I think it's more standard to read from an environment variable or a system property. Then there's less indirection.

.circleci/config.yml
karenfeng: PySpark isn't released in the nightly snapshots. If we wanted to test against PySpark 3.0, we'd have to use one of the preview releases (eg. https://archive.apache.org/dist/spark/spark-3.0.0-preview2/), which we are currently ahead of.

.circleci/config.yml
karenfeng: Sure, I made the change.

.circleci/config.yml
```diff
@@ -131,10 +131,58 @@ jobs:
...
+      - restore_cache:
+          keys:
+            - conda-deps-v1-{{ checksum "python/environment.yml" }}
+      - *install_deps
+      - run:
+          name: Use latest PySpark release
+          environment:
+          command: |
+            export PATH=$HOME/conda/envs/glow/bin:$PATH
+            wget https://archive.apache.org/dist/spark/spark-3.0.0-preview2/pyspark-3.0.0.dev2.tar.gz
```
henrydavidge, line 147: Doesn't this mess up the cached conda environment?

.circleci/config.yml
```diff
@@ -131,10 +131,58 @@ jobs:
...
+            sbt docs_2_12/test exit
+
 workflows:
   version: 2
   test:
     jobs:
       - check-links
       - scala-2_11-tests
       - scala-2_12-tests
+      - spark-3-tests
```
henrydavidge, line 178: So we want to run this build against PRs and nightly?

Hoeze:
Is there already a working build for Spark 3.0-rc1?

karenfeng:
@Hoeze, our build is compatible with the nightly Spark 3 snapshots; this might be ahead of rc1. We don't plan to release Glow on Spark 3 until after the public release, but in the meantime you can build a fat jar locally with https://github.com/sbt/sbt-assembly.

build.sbt
```diff
@@ -1,15 +1,21 @@
...
 
-val sparkVersionMajorMinor = "2.4"
-val sparkVersion = "2.4.3"
-
 lazy val scala212 = "2.12.8"
 lazy val scala211 = "2.11.12"
 
+lazy val sparkVersion = sys.env.getOrElse("SPARK_VERSION", "2.4.3")
+
+lazy val minorVersion = (version: String) =>
```
henrydavidge, line 13: nit: this should still be called `majorMinor`. Also, why is it a lazy val pointing to an anonymous function? Can it just be a lazy val, or at least a def?

PR #182, opened 2020-03-31
[HLS-387] GFF3 Data Source

kianfar77:

This PR adds the "gff" datasource to glow for reading GFF3 (https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md) files. The usage will be like:
``val df = spark.read.format("gff").load``  

The data source is able to infer the schema or accept a user-specified schema. It flattens the attributes field by creating a column for each tag that appears in the attributes column of the gff file.

The inferred schema will have base fields corresponding to the first 8 columns of gff3 called seqId, source, type, start, end, score, strand, and phase, followed by any official attribute field among id, name, alias, parent, target, gap, derivesfrom, note, dbxref, ontologyterm, and iscircular that appears in the gff tags followed by any unofficial attribute field that appears in the tags. In the inferred schema, the base and official fields will be in the same order as listed above. The unofficial fields will be in alphabetical order.

Any user-specified schema can have any subset of fields corresponding to the 9 columns of gff3 (named seqId, source, type, start, end, score, strand, phase, and attributes), the official attribute fields, and the unofficial attribute fields. The name of the official and unofficial fields should match the tag name in a case-and-underscore-insensitive fashion.

## How is this patch tested?
- [x] Unit tests
- [x] Integration tests
- [x] Manual tests

(Details)


core/src/main/scala/io/projectglow/gff/GffFileFormat.scala
```diff
@@ -0,0 +1,328 @@
...
+      columnPruning = sparkSession.sessionState.conf.csvColumnPruning,
+      sparkSession.sessionState.conf.sessionLocalTimeZone
+    )
+
+    val csv = GffFileFormat
+      .createBaseDataset(sparkSession, files, parsedOptions)
+      .select(attributesField.name)
+      .filter(col(attributesField.name).isNotNull)
+
+    // Use of ParsedAttributesToken helps update sep as soon as detected and avoid looking into
```
kianfar77, line 82: Auto-detection of tag-value separator with spark sql functions is not straightforward.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,245 @@
...
+  /**
+   * Creates relation
+   *
+   * @param sqlContext spark sql context
+   * @param parameters parameters for job
+   * @return Base relation
+   */
+  override def createRelation(sqlContext: SQLContext, parameters: Map[String, String]): BaseRelation = {
+
+    createRelation(sqlContext, parameters, null)
```
henrydavidge, line 45: Think it would be better to infer schema here and then call the method below.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+}
+
+
+class GffResourceRelation(
+  val sqlContext: SQLContext,
+  path: String,
+  requiredSchema: StructType
+) extends BaseRelation with TableScan  { // TODO: investigate use of PrunedFilteredScan
```
henrydavidge, line 78: IMO the column pruning would be helpful and probably very simple to implement. I doubt the filters are worth spending time on.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,245 @@
...
+        )
+      )
+      .collect()(0).getAs[Seq[String]](0)
+
+    // generate the schema. The field names will be case and underscore insensitive
+    val officialAttributeFields =
+      gffOfficialAttributeFields.foldLeft(Seq[StructField]()) { (s, f) =>
+        if (attributeTags
+          .map(t => deUnderscore(t.toLowerCase))
+          .contains(f.name.toLowerCase)) {
```
henrydavidge, line 172: FYI, I think that spark sql is already case insensitive

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    StructType(
+      gffBaseSchema.fields ++ officialAttributeFields ++ unofficialAttributeFields
+    )
+
+  }
+
+  def filterCommentsAndFasta(df: DataFrame): DataFrame = {
+    df.where(
```
henrydavidge, line 198: There's an option in the CSV reader to drop rows that don't have the right number of fields. Might save you some code here.

henrydavidge: As well as an option to filter out comments.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,245 @@
...
+        )
+      )
+      .agg(
+        collect_set(
+          attributesMapColumnName
+        )
+      )
+      .collect()(0).getAs[Seq[String]](0)
+
+    // generate the schema. The field names will be case and underscore insensitive
```
henrydavidge, line 167: Code style thing: I think this would be more concise and clear if you normalized all the names together and then used [sortBy](https://www.scala-lang.org/api/2.12.0/scala/collection/Seq.html#sortBy[B](f:A=%3EB)(implicitord:scala.math.Ordering[B]):Repr) to put them in the order you want.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,245 @@
...
+          )
+        case f => // TODO: To be completed (also rectify column names)
+          df.withColumn(
+            f.name,
+            element_at(col(attributesMapColumnName), f.name)
+          )
+      }
+    )
+      .drop(attributesField.name, attributesMapColumnName) // TODO: Complete the option of keeping attributes field
+      .rdd
```
henrydavidge, line 131: For an easy performance gain, we can override [needConversion](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/sources/interfaces.scala#L211-L223) and return `queryExecution.toRdd`.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,245 @@
...
+
+  def deUnderscore(s: String): String = {
+    s.replaceAll("_", "")
+  }
+
+  private[gff] val gffCsvSchema = StructType(
+    gffBaseSchema
+      .fields
+      .toSeq
+      .map(f => StructField(f.name, StringType))
```
henrydavidge, line 229: Why do you have to parse everything as a string at first? You can't parse them as the proper data types?

kianfar77:
> I had some preliminary comments but it looks like a great start to me!
> 
> One question: are bgzip files with a `.gz` extensions handled correctly now? I'm not sure if we'll know they're splittable.

As we discussed, it is handled but not in splittable fashion. Will address in a separate PR by extending CSV datasource as we discussed.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Done

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Incorporated ColumnPruning. Skipped Filters

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: True. The case sensitivity against tag names needs to be dealt with though. Revise this section based on your other comments.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Used filtering the comments. 
Used DROPMALFORMED. This has issues in spark 2.4.4 and before which I dealt with and documented in the code.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Revised.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Great. Done

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: I used nullValue option of csv reader and was able to do it.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,346 @@
...
+      )
+      .collect()(0)
+      .getAs[Seq[String]](0)
+      .filter(!_.isEmpty)
+
+    spark.conf.set(columnPruningConf, originalColumnPruning)
+
+    val attributeFields = attributeTags
+      .foldLeft(Seq[StructField]()) { (s, t) =>
+        val officialIdx = officialFieldNames.indexOf(normalizeString(t))
```
henrydavidge, line 276: nit: might be a little cleaner to `gffOfficialAttributeFields.find(f => f.name == normalizeString(t)).map(_.dataType).getOrElse(StringType)`

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,346 @@
...
+        if (officialIdx > -1) {
+          s :+ StructField(t, gffOfficialAttributeFields(officialIdx).dataType)
+        } else {
+          s :+ StructField(t, StringType)
+        }
+      }
+      .sortBy(_.name)(FieldNameOrdering)
+
+    StructType(
+      gffBaseSchema.fields.dropRight(1) ++ attributeFields
```
henrydavidge, line 286: nit: comment that `attributes` is the last field in `gffBaseSchema.fields`

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,346 @@
...
+            |       k -> regexp_replace(lower(k), "_", "")
+            |   )""".stripMargin
+        ),
+        map_values(col(attributesMapColumnName))
+      )
+    )
+  }
+
+  def filterFastaLines(df: DataFrame): DataFrame = {
+    df.where(
```
henrydavidge, line 319: This is kind of a confusing way to write this filter. `fold` is specifically called out in our style guide as a hard-to-understand pattern: https://github.com/databricks/scala-style-guide. [Coalesce](https://spark.apache.org/docs/2.4.0/api/scala/index.html#org.apache.spark.sql.functions$@coalesce(e:org.apache.spark.sql.Column*):org.apache.spark.sql.Column) might be a good alternative here.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,346 @@
...
+    val attributeFields = attributeTags
+      .foldLeft(Seq[StructField]()) { (s, t) =>
+        val officialIdx = officialFieldNames.indexOf(normalizeString(t))
+        if (officialIdx > -1) {
+          s :+ StructField(t, gffOfficialAttributeFields(officialIdx).dataType)
+        } else {
+          s :+ StructField(t, StringType)
+        }
+      }
+      .sortBy(_.name)(FieldNameOrdering)
```
henrydavidge, line 283: nit: I think the custom ordering is not necessary. It might be simpler to use the natural ordering on whether the field is in the official attributes and then the name.

```scala
attributes.sortBy(field => (!officialAttributes.contains(field), field.name))
```

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,346 @@
...
+        )
+      )
+      .collect()(0)
+      .getAs[Seq[String]](0)
+      .filter(!_.isEmpty)
+
+    spark.conf.set(columnPruningConf, originalColumnPruning)
+
+    val attributeFields = attributeTags
+      .foldLeft(Seq[StructField]()) { (s, t) =>
```
henrydavidge, line 275: Why is this a `fold` instead of a simple `map`?

core/src/test/scala/io/projectglow/gff/GffReaderSuite.scala
```diff
@@ -0,0 +1,249 @@
...
+    :+ StructField("is_Circular", BooleanType)
+  )
+
+  private val testSchemaExpected = StructType(
+    gffBaseSchema.fields :+
+    StructField("Dbxref", ArrayType(StringType))
+    :+ StructField("Is_circular", BooleanType)
+  )
+
+  test("Schema inference") {
```
henrydavidge, line 81: Are there any particular edge cases we should look for here? E.g., no attributes fields, attributes fields containing weird characters (can there be escaped `=` or `;` inside attribute values)?

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
```diff
@@ -0,0 +1,346 @@
...
+   * Creates relation
+   *
+   * @param sqlContext spark sql context
+   * @param parameters parameters for job
+   * @return Base relation
+   */
+  override def createRelation(
+      sqlContext: SQLContext,
+      parameters: Map[String, String]): BaseRelation = {
+    val path = parameters.get("path")
```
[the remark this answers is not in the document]
kianfar77: Done

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Done

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Done

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Used Coalesce

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: Used your suggestion but with indexOf because I want the official fields to be in a set order as well.

core/src/main/scala/io/projectglow/gff/GffDataSource.scala
kianfar77: remnant of the older version! didnt pay attention that does nothing. Changed.

kianfar77:
> Made a first pass. I think the high level is really solid! Most of my comments are about code style and readability.
> 
> I'm a little surprised there aren't more edge cases to look at. We should be sure we've thought hard about the kinds of "weird" GFFs we'll encounter.

Please see my comment in the tests. GFF3 format is pretty restricted. I could not think of cases other than those mentioned that are both weird and valid.

core/src/test/scala/io/projectglow/gff/GffReaderSuite.scala
```diff
@@ -0,0 +1,280 @@
...
+        null,
+        "true",
+        "passed",
+        null
+      )
+    )
+    assert(dfRows.sameElements(expectedRows))
+  }
+
+  test("Read gff with user-specified schema containing attributesField") {
```
henrydavidge, line 192: What's the expected behavior if the user provides a schema with fields that are not actually in the attributes map? We should have a test for that.

core/src/test/scala/io/projectglow/gff/GffReaderSuite.scala
```diff
@@ -0,0 +1,280 @@
...
+    val df = spark
+      .read
+      .format(sourceName)
+      .load(testGff3)
+    val e = intercept[UnsupportedOperationException] {
+      df.write.format(sourceName).save("noop")
+    }
+    assert(e.getMessage.contains("GFF data source does not support writing!"))
+  }
+
```
henrydavidge, line 279: We should have a test for column pruning where we select only a few columns from the inferred schema.

kianfar77: Added the "column pruning" test

core/src/test/scala/io/projectglow/gff/GffReaderSuite.scala
kianfar77: The column will be added. All entries will be null. Added a test for that.

PR #184, opened 2020-04-06
Improve errors when user attempts to save unwritable data to VCF

henrydavidge:
Previously, if a user tried to a write a DataFrame that was missing required fields like `contigName`, they could get confusing errors that they were trying to write an empty DF. Since our default stringency is "silent" (perhaps a bad decision!), we'd drop VCs if they failed conversion. Now, before writing we check the schema to make sure that required fields are present. In addition, I cleaned up the behavior for "strict" stringency so that conversion errors cause the write to fail.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/vcf/InternalRowToVariantContextConverter.scala
+++ b/core/src/main/scala/io/projectglow/vcf/InternalRowToVariantContextConverter.scala
@@ -166,7 +166,9 @@ class InternalRowToVariantContextConverter(
       try {
         Option(builder.make())
       } catch {
-        case NonFatal(ex) =>
+        case NonFatal(_) if stringency == ValidationStringency.SILENT =>
+          None
+        case NonFatal(ex) if stringency == ValidationStringency.LENIENT =>
           provideWarning(s"Could not build variant context: ${ex.getMessage}")
           None
       }
--- a/core/src/main/scala/io/projectglow/vcf/InternalRowToVariantContextConverter.scala
+++ b/core/src/main/scala/io/projectglow/vcf/InternalRowToVariantContextConverter.scala
@@ -166,9 +166,7 @@ class InternalRowToVariantContextConverter(
       try {
         Option(builder.make())
       } catch {
-        case NonFatal(_) if stringency == ValidationStringency.SILENT =>
-          None
-        case NonFatal(ex) if stringency == ValidationStringency.LENIENT =>
+        case NonFatal(ex) =>
           provideWarning(s"Could not build variant context: ${ex.getMessage}")
           None
       }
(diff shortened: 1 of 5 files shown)


core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
```diff
@@ -273,6 +273,23 @@ object VCFFileFormat {
       false
     }
   }
+
+  def requireWritableAsVCF(schema: StructType): Unit = {
+    val baseRequiredFields =
+      Seq(VariantSchemas.contigNameField, VariantSchemas.startField, VariantSchemas.refAlleleField)
```
karenfeng, line 279: I believe `endField` is also required as well. To check this more thoroughly, we could have tests that remove each of the columns in a DataFrame and attempt conversion to a `VariantContext`. If conversion fails, we should verify that the check for writable-ness also fails, and vice versa.

core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
```diff
@@ -273,6 +273,23 @@ object VCFFileFormat {
...
+    val requiredFields = if (schema.exists(_.name == VariantSchemas.genotypesFieldName)) {
+      baseRequiredFields :+ VariantSchemas.alternateAllelesField
```
karenfeng, line 281: Is this strictly true? It looks like we can successfully make `Genotype`s as long as the appropriate allele exists in the `VariantContext` (i.e. only called reference allele).

core/src/test/scala/io/projectglow/vcf/VCFFileWriterSuite.scala
```diff
@@ -400,6 +400,31 @@ abstract class VCFFileWriterSuite(val sourceName: String)
     assert(sampleIds.length == 1)
     assert(sampleIds.head == Seq("sample_1", "sample_2", "sample_3"))
   }
+
+  test("validate schema before write") {
```
karenfeng, line 404: This would benefit from more thorough testing - at minimum, should iterate through each of the expected required fields. Ideally, we should have a sanity check to make sure that our expected required fields match the actual required fields.

core/src/main/scala/io/projectglow/vcf/InternalRowToVariantContextConverter.scala
```diff
@@ -166,7 +166,9 @@ class InternalRowToVariantContextConverter(
       try {
         Option(builder.make())
       } catch {
-        case NonFatal(ex) =>
+        case NonFatal(_) if stringency == ValidationStringency.SILENT =>
+          None
+        case NonFatal(ex) if stringency == ValidationStringency.LENIENT =>
```
karenfeng, line 171: I believe we already do this behavior with `provideWarning`.

core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
henrydavidge: As discussed today, I think it's better to disallow this. It's possible to form a variant context, but unlikely to result in the behavior users expect. They can always add a dummy `alternateAlleles` if it's really what they want.

core/src/test/scala/io/projectglow/vcf/VCFFileWriterSuite.scala
henrydavidge: Done

PR #189, opened 2020-04-14
Do not read contig and filter header lines on VCF ingest

karenfeng:

With https://github.com/projectglow/glow/pull/173, VCF ingest now fails when merging files with contig lines that are missing the `length` field. This is resolved in newer versions of the HTSJDK as of https://github.com/samtools/htsjdk/pull/1418.

As we don't need contig and filter header lines for schema inference during VCF ingest, we can skip reading them entirely. This PR adds an option to skip reading non-schema header lines while reading unique header lines.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

### DIFF
--- a/core/src/main/scala/io/projectglow/vcf/VCFHeaderUtils.scala
+++ b/core/src/main/scala/io/projectglow/vcf/VCFHeaderUtils.scala
@@ -103,19 +103,23 @@ object VCFHeaderUtils extends GlowLogging {
   }
 
   /**
-   * Find the unique header lines from an RDD of VCF headers.
+   * Find the unique desired header lines from an RDD of VCF headers.
    * If lines of the same class found with the same ID, we pick one unless they are incompatible.
    * If there are incompatible lines, [[IllegalArgumentException]] is thrown.
    * Incompatible lines are:
    * - FORMAT or INFO lines with the same ID but different types or counts
    * - contig lines with the same ID but different lengths
    */
-  def getUniqueHeaderLines(headers: RDD[VCFHeader]): Seq[VCFHeaderLine] = {
+  def getUniqueHeaderLines(
+      headers: RDD[VCFHeader],
+      getNonSchemaHeaderLines: Boolean): Seq[VCFHeaderLine] = {
     headers.flatMap { header =>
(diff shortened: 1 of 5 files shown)


core/src/main/scala/io/projectglow/vcf/VCFHeaderUtils.scala
```diff
@@ -103,19 +103,23 @@ object VCFHeaderUtils extends GlowLogging {
...
-  def getUniqueHeaderLines(headers: RDD[VCFHeader]): Seq[VCFHeaderLine] = {
+  def getUniqueHeaderLines(
+      headers: RDD[VCFHeader],
+      getNonSchemaHeaderLines: Boolean): Seq[VCFHeaderLine] = {
     headers.flatMap { header =>
       val infoHeaderLines = header.getInfoHeaderLines.asScala
       val formatHeaderLines = header.getFormatHeaderLines.asScala
-      val contigHeaderLines = header.getContigLines.asScala
-      val filterHeaderLines = header.getFilterLines.asScala
+      val contigHeaderLines =
```
henrydavidge, line 119: nit: can you consolidate into a single `if else`? Generally it's less error-prone to check flags in as few places as possible.

core/src/main/scala/io/projectglow/vcf/VCFHeaderUtils.scala
```diff
@@ -103,19 +103,23 @@ object VCFHeaderUtils extends GlowLogging {
   }
 
   /**
-   * Find the unique header lines from an RDD of VCF headers.
+   * Find the unique desired header lines from an RDD of VCF headers.
    * If lines of the same class found with the same ID, we pick one unless they are incompatible.
    * If there are incompatible lines, [[IllegalArgumentException]] is thrown.
    * Incompatible lines are:
    * - FORMAT or INFO lines with the same ID but different types or counts
    * - contig lines with the same ID but different lengths
```
henrydavidge, line 111: Add a brief comment about the meaning of `getNonSchemaHeaderLines`.

PR #193, opened 2020-04-24
Fix expression encoders for Spark 3.0

hvanhovell:
In order to make ExpressionEncoders immutable and thread safe the `fromRow` and `toRow` functions were removed, and the `createSerializer` and `createDeserializer` methods were added in their place. See https://github.com/apache/spark/commit/e7fef70fbbea08a38316abdaa9445123bb8c39e2 for more information.

This updates Glue to use these new methods for Spark 3.0. 

## How is this patch tested?
- [x] Unit tests


hvanhovell:
cc @henrydavidge @karenfeng

core/src/main/shim/2.4/SparkShim.scala
```diff
@@ -51,6 +52,14 @@ object SparkShim extends SparkShimBase {
     )
   }
 
+  override def createSerializer[T](encoder: ExpressionEncoder[T]): T => InternalRow = {
+    encoder.copy().toRow _
```
henrydavidge, line 56: nit: just `toRow` or `toRow(_)`. We don't typically use infix notation.

PR #194, opened 2020-04-29
[HLS-425] Remove dependency to GATK's SimpleInterval

kianfar77:
This PR replaces the dependency to ``org.broadinstitute.hellbender.utils.SimpleInterval`` in Glow with an internal SimpleInterval class. 

## How is this patch tested?
- [x] Unit tests
- [x] Integration tests
- [x] Manual tests

(Details)


core/src/main/scala/io/projectglow/common/GenomicIntervalUtils.scala
```diff
@@ -0,0 +1,156 @@
...
+package io.projectglow.common
+
+import scala.math.{max, min}
+
+object GenomicIntervalUtils {
+
+  /**
+   * A class representing a closed interval of longs that also accommodates an empty interval.
+   */
+  class LongInterval() extends Serializable {
```
henrydavidge, line 26: It seems like this should just be a
```scala
case class LongInterval(coordinates: Option[(Long, Long)])
```

Then you get `isSame` (which should just be `equals`), `isEmpty`, and `nonEmpty` for free.

core/src/main/scala/io/projectglow/common/GenomicIntervalUtils.scala
```diff
@@ -0,0 +1,156 @@
...
+    var interval: Option[(Long, Long)] = None
+
+    def this(start: Long, end: Long) {
+      this()
+      if (end >= start) {
+        this.interval = Option((start, end))
+      }
+    }
+
+    def start: Long =
```
henrydavidge, line 36: IMO it's bad practice for simple accessors like this to throw exceptions. It's better to just expose method that gets the optional start. I think that the simple `coordinates` covers that.

core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
```diff
@@ -27,151 +27,52 @@ import org.apache.hadoop.conf.Configuration
...
-        i.getContig == j.getContig && i.getStart == j.getStart && i.getEnd == j.getEnd
-      case (None, _) => other.isEmpty
-      case (_, None) => interval.isEmpty
+case class ParsedFilterResult(
+    contig: Contig,
+    startInterval: LongInterval,
+    endInterval: LongInterval)
+
+case class ContigAndInterval(contig: Contig, interval: LongInterval) extends Serializable {
+  def toHTSJDKInterval: Interval = {
```
henrydavidge, line 41: This should return an option.

kianfar77:
Based on your comments, I think you prefer minimal change to existing code. Revised for minimal change.

PR #200, opened 2020-04-30
Install pyarrow in dev environment; use conda for more dependencies

henrydavidge:

Since we're going to add code that uses pandas UDFs, we need to install `pyarrow`. 

Also, I moved more dependencies to conda instead of pip. Since we're using conda as an environment manager, it's valuable to install as many dependencies as possible through conda, particularly because there is limited dependency resolution between packages installed by conda and those install by pip.

## How is this patch tested?
- [ ] Unit tests
- [x] Integration tests
- [x] Manual tests

(Details)


.circleci/config.yml
```diff
@@ -24,6 +24,7 @@ install_pyspark3: &install_pyspark3
       export PATH=$HOME/conda/bin:$PATH
       if [ ! -e "pyspark-3.0.0.dev2.tar.gz" ]; then
         source activate glow
+        conda remove -y pyspark
```
karenfeng, line 27: Good idea :)

python/environment.yml
```diff
@@ -4,16 +4,18 @@ channels:
 dependencies:
   - python=3.7
   - jinja2
+  - numpy=1.17.4
+  - nomkl # Skip MKL for local development
```
karenfeng, line 8: What installs MKL by default?

henrydavidge: numpy

PR #205, opened 2020-05-06
Remove splitToBiallelic option from VCF reader doc page

kianfar77:

This PR removes the splitToBiallelic from the VCF reader doc page and updates the embedded notebook to remove use of that option. A note referring the user to split_multiallelics transformer is added to the page. A few remnants of splitToBiallelic in unit tests are removed as well.

## How is this patch tested?
- [x] Unit tests
- [x] Integration tests
- [X] Manual tests

(Details)


kianfar77:
Changed ``splitToBiallelic`` in panda-lmm and gwas notebooks to split transformer

PR #206, opened 2020-05-07
Remove ExpressionEncoder

karenfeng:

Removes use of the non-public Spark API `ExpressionEncoder` in Glow to avoid breaking with API changes.

This PR also removes the private `vcfRowSchema` schema option as well as the accompanying converters.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


core/src/test/scala/io/projectglow/vcf/VCFRowToVariantContextConverterSuite.scala
```diff
@@ -30,14 +30,23 @@ import org.apache.commons.io.FileUtils
 import io.projectglow.common.{GenotypeFields, VCFRow}
 import io.projectglow.sql.GlowBaseTest
 
-class VCFRowToVariantContextConverterSuite extends GlowBaseTest with VCFConverterBaseTest {
+class VCFRowToVariantContextConverterSuite extends VCFConverterBaseTest {
```
henrydavidge, line 33: Seems like we should rename this to `InternalRowToVariantContextConverter`

core/src/main/scala/io/projectglow/sql/expressions/LogisticRegressionGwas.scala
```diff
@@ -43,6 +43,10 @@ case class LogitTestResults(
 
 object LogitTestResults {
   val nanRow: InternalRow = InternalRow(NaN, NaN, ArrayData.toArrayData(Seq(NaN, NaN)), NaN)
+  lazy val schema: StructType = ScalaReflection
```
henrydavidge, line 46: Is this only used in test code? If so please move to the test package.

henrydavidge: Btw, I don't think it's really any better to use `ScalaReflection` than `ExpressionEncoder`. It's an equally private interface.

karenfeng: This seems to be publicly documented: https://kb.databricks.com/data/schema-from-case-class.html

PR #207, opened 2020-05-08
Create mean substitution function

karenfeng:

Creates a function to substitute the missing values of a numeric array using the mean of the non-missing values. Values that are `NaN`, null, or equal to the optional missing value parameter (which defaults to `-1` if not provided) are considered missing.

This is useful for GWAS use-cases; after calling `genotype_states`, `mean_substitute` can be used to impute the missing values.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


python/glow/functions.py
```diff
@@ -315,6 +315,36 @@ def normalize_variant(contigName: Union[Column, str], start: Union[Column, str],
     assert check_return_type(output)
     return output
 
+
+def impute_mean(array: Union[Column, str], missingValue: Union[Column, str] = None) -> Column:
```
henrydavidge, line 319: From a grammatical perspective, we impute the missing values, not the array or the mean(cf [the scipy docs](https://scikit-learn.org/stable/modules/impute.html)). So I think that a better name would be `impute_by_mean` or `impute_missing` or maybe just `impute` for shorthand. If we expect imputation by mean to be way more common than other methods (I think this is true?), I think it would make sense to have a string typed "strategy" argument that is `mean` by default.

core/src/main/scala/io/projectglow/sql/expressions/ImputeMean.scala
```diff
@@ -0,0 +1,100 @@
...
+  override def children: Seq[Expression] = Seq(array, missingValue)
+
+  def this(array: Expression) = {
+    this(array, Literal(-1))
+  }
+
+  override def rewrite: Expression = {
+    val nLv = NamedLambdaVariable(
+      "numArg",
+      array.dataType.asInstanceOf[ArrayType].elementType,
```
henrydavidge, line 45: First require that it's an array type so that the error message makes sense.

core/src/main/scala/io/projectglow/sql/expressions/ImputeMean.scala
```diff
@@ -0,0 +1,100 @@
...
+      true
+    )
+    val sLv = NamedLambdaVariable("structArg", StructType.fromDDL("sum double, count long"), true)
+
+    val isMissing = Or(IsNaN(nLv), Or(IsNull(nLv), EqualTo(nLv, missingValue)))
```
henrydavidge, line 50: You could make this code more readable by using the DSL in `org/apache/spark/sql/catalyst/dsl/package.scala`. It might make sense to add something to an internal dsl for higher order functions.

core/src/main/scala/io/projectglow/sql/expressions/ImputeMean.scala
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.spark.sql.types.{ArrayType, DataType, DoubleType, LongType, StringType, StructType}
+import org.apache.spark.unsafe.types.UTF8String
+
+/**
+ * Imputes the missing values of an array using the mean of the non-missing values. Values that are NaN, null or
+ * equal to the missing value parameter are not included in the aggregation, and are substituted with the mean.
+ *
+ * If the missing value is not provided, the parameter defaults to -1.
+ */
+case class ImputeMean(array: Expression, missingValue: Expression) extends RewriteAfterResolution {
```
henrydavidge, line 35: What happens if the array element is not numeric? Or if it's numeric but not double? Might need to `ImplicitCastInputTypes` to doubles.

core/src/main/scala/io/projectglow/sql/expressions/ImputeMean.scala
karenfeng: Do you have any suggestions for a simpler way to express higher-order functions?

functions.yml
```diff
@@ -244,6 +244,29 @@ etl:
...
+          >>> df.select(glow.impute('unimputed_values', lit('mean'), lit(0.0)).alias('imputed_values')).collect()
+          [Row(imputed_values=[2.5, 2.5, 2.5, 1.0, 2.0, 3.0, 4.0])]
+          >>> df = spark.createDataFrame([Row(unimputed_values=[0, 1, 2, 3, -1, None])])
+          >>> df.select(glow.impute('unimputed_values', lit('mean')).alias('imputed_values')).collect()
+          [Row(imputed_values=[0.0, 1.0, 2.0, 3.0, 1.5, 1.5])]
+      args:
+        - name: array
+          doc: The array of values to impute
+        - name: strategy
+          doc: The imputation strategy. Currently supports only ``mean``.
```
henrydavidge, line 265: IMO we should default to `mean`

henrydavidge: For genetics, I'd imagine that people would use a different missing value more often than median imputation

core/src/main/scala/io/projectglow/sql/expressions/Impute.scala
```diff
@@ -0,0 +1,171 @@
...
+      NamedLambdaVariable("numArg", array.dataType.asInstanceOf[ArrayType].elementType, true)
+    val sLv = NamedLambdaVariable("structArg", StructType.fromDDL("sum double, count long"), true)
+    val sumName = Literal(UTF8String.fromString("sum"), StringType)
+    val countName = Literal(UTF8String.fromString("count"), StringType)
+    val getSum = GetStructField(sLv, 0, Some("sum"))
+    val getCount = GetStructField(sLv, 1, Some("count"))
+
+    // Sum and count of non-missing values
+    val zeros = namedStruct(sumName, Literal(0d), countName, Literal(0L))
+    // Update sum and count with array value
```
henrydavidge, line 68: We should use an algorithm that's not as likely to overflow, for example https://stackoverflow.com/a/1346890

karenfeng: I'm not sure if we're concerned about overflow in the common case. Given that we're generally working with genotype states (0, 1, or 2) and supporting <10M individuals, I don't think we'd hit the maximum value of a Double (~1.8E308). Calculating the mean via this method will require (N-1) additional divisions and (potentially but less importantly) introduce increased floating point error.

functions.yml
karenfeng: I'll remove support for other strategies for now.

core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
```diff
@@ -0,0 +1,118 @@
...
+      zeros,
+      updateFn,
+      finalizeFn
+    )
+  }
+
+  override def rewrite: Expression = {
+    if (!array.dataType.isInstanceOf[ArrayType] ||
+      !array.dataType.asInstanceOf[ArrayType].elementType.isInstanceOf[NumericType]) {
+      throw new IllegalArgumentException(
```
henrydavidge, line 98: nit: I think we usually throw `AnalysisException`s for errors during analysis.

karenfeng: Unfortunately I believe the constructor for AnalysisException is package-private: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/AnalysisException.scala

PR #209, opened 2020-05-13
Use full ExpressionInfo constructor

karenfeng:

Use the full ExpressionInfo constructor.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/core/src/main/shim/3.0/SparkShim.scala
+++ b/core/src/main/shim/3.0/SparkShim.scala
@@ -30,6 +30,8 @@ object SparkShim extends SparkShimBase {
     parser.parse(input)
   }
 
+  // [SPARK-31429][SQL][DOC] Automatically generates a SQL document for built-in functions
+  // Adds 'group' argument to the ExpressionInfo constructor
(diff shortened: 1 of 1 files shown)


core/src/main/shim/3.0/SparkShim.scala
```diff
@@ -41,13 +43,17 @@ object SparkShim extends SparkShimBase {
...
     new ExpressionInfo(
       className,
       db,
       name,
       usage,
-      arguments
+      arguments,
+      examples,
+      note,
+      "", // group
```
henrydavidge, line 54: We have `group` defined in `functions.yml`, can we add it here?

karenfeng: The constructor verifies that `group` fall into one of Spark's valid group categories (https://github.com/apache/spark/blob/c8f3bd861d96cf3f7b01cd9f864c181a57e1c77a/sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/expressions/ExpressionInfo.java#L139), which don't match with our groups.

PR #211, opened 2020-05-14
Fix mean substitution lambda variable usage

karenfeng:

The mean substitution expression reused lambda variables in an unsafe way.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual tests

### DIFF
--- a/core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
+++ b/core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
@@ -47,24 +47,26 @@ case class MeanSubstitute(array: Expression, missingValue: Expression)
   // A value is considered missing if it is NaN, null or equal to the missing value parameter
   def isMissing(lv: NamedLambdaVariable): Predicate = IsNaN(lv) || IsNull(lv) || lv === missingValue
 
-  def getNamedStruct(sumValue: Expression, countValue: Expression): Expression = {
+  def createNamedStruct(sumValue: Expression, countValue: Expression): Expression = {
     val sumName = Literal(UTF8String.fromString("sum"), StringType)
     val countName = Literal(UTF8String.fromString("count"), StringType)
     namedStruct(sumName, sumValue, countName, countValue)
   }
 
-  def getNumericLambdaVariable(): NamedLambdaVariable =
+  def createNumericLambdaVariable(): NamedLambdaVariable =
     NamedLambdaVariable("numArg", array.dataType.asInstanceOf[ArrayType].elementType, true)
-  def getStructLambdaVariable(): NamedLambdaVariable =
+  def createStructLambdaVariable(): NamedLambdaVariable =
     NamedLambdaVariable("structArg", StructType.fromDDL("sum double, count long"), true)
 
-  def getSum(lv: NamedLambdaVariable): GetStructField = GetStructField(lv, 0, Some("sum"))
-  def getCount(lv: NamedLambdaVariable): GetStructField = GetStructField(lv, 1, Some("count"))
+  def getSum(stateStruct: NamedLambdaVariable): GetStructField =
+    GetStructField(stateStruct, 0, Some("sum"))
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
```diff
@@ -47,42 +47,59 @@ case class MeanSubstitute(array: Expression, missingValue: Expression)
   // A value is considered missing if it is NaN, null or equal to the missing value parameter
   def isMissing(lv: NamedLambdaVariable): Predicate = IsNaN(lv) || IsNull(lv) || lv === missingValue
 
-  def getMean: Expression = {
-    val nLv =
-      NamedLambdaVariable("numArg", array.dataType.asInstanceOf[ArrayType].elementType, true)
-    val sLv = NamedLambdaVariable("structArg", StructType.fromDDL("sum double, count long"), true)
+  def getNamedStruct(sumValue: Expression, countValue: Expression): Expression = {
```
henrydavidge, line 50: `createNamedStruct`

core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
```diff
@@ -47,42 +47,59 @@ case class MeanSubstitute(array: Expression, missingValue: Expression)
...
-    val getSum = GetStructField(sLv, 0, Some("sum"))
-    val getCount = GetStructField(sLv, 1, Some("count"))
+    namedStruct(sumName, sumValue, countName, countValue)
+  }
 
-    // Sum and count of non-missing values
-    val zeros = namedStruct(sumName, Literal(0d), countName, Literal(0L))
-    // Update sum and count with array value
-    val updateFn = LambdaFunction(
+  def getNumericLambdaVariable(): NamedLambdaVariable =
```
henrydavidge, line 56: `create`

core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
```diff
@@ -47,42 +47,59 @@ case class MeanSubstitute(array: Expression, missingValue: Expression)
...
+    NamedLambdaVariable("numArg", array.dataType.asInstanceOf[ArrayType].elementType, true)
+  def getStructLambdaVariable(): NamedLambdaVariable =
+    NamedLambdaVariable("structArg", StructType.fromDDL("sum double, count long"), true)
+
+  def getSum(lv: NamedLambdaVariable): GetStructField = GetStructField(lv, 0, Some("sum"))
```
henrydavidge, line 61: Think it would be more clear to call the argument `stateStruct` or something like that.

core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
```diff
@@ -47,42 +47,59 @@ case class MeanSubstitute(array: Expression, missingValue: Expression)
...
         )
       ),
       Seq(sLv, nLv)
     )
-    // Calculate value for imputation
-    val finalizeFn = LambdaFunction(
+  }
+
+  // Calculate mean for imputation
+  def finalizeFn: LambdaFunction = {
```
henrydavidge, line 85: Since we can only use this once, I think it would be more clear as a `lazy val`

core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
```diff
@@ -47,42 +47,59 @@ case class MeanSubstitute(array: Expression, missingValue: Expression)
...
-        getSum / getCount,
+        getSum(sLv) / getCount(sLv),
         // If all values were missing, substitute with missing value
         missingValue
       ),
       Seq(sLv)
     )
+  }
+
+  def getMean: Expression = {
```
henrydavidge, line 100: `lazy val arrayMean`

PR #213, opened 2020-05-16
Support numpy literals

karenfeng:

We use `DenseMatrix` in our regression methods, but this is not a commonly used format to store matrix data in Python. This PR introduces support for numpy literals from double-typed numpy ndarrays under the hood with py4j once `glow` is imported. With `lit()`:
- For 1-dimensional arrays, the conversion is to Java arrays. (This allows the user to bypass the conversions to Python lists and Java lists.)
- For 2-dimensional arrays, the conversion is to DenseMatrix.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

Python tests.


python/glow/conversions.py
```diff
@@ -0,0 +1,91 @@
...
+        java_arr = _convert_numpy_to_java_array(flat_arr)
+        dense_matrix = sc._jvm.org.apache.spark.ml.linalg.DenseMatrix(object.shape[0], object.shape[1], java_arr)
+        matrix_udt = sc._jvm.org.apache.spark.ml.linalg.MatrixUDT()
+        converter = sc._jvm.org.apache.spark.sql.catalyst.CatalystTypeConverters.createToCatalystConverter(matrix_udt)
+        literal_matrix = sc._jvm.org.apache.spark.sql.catalyst.expressions.Literal.create(
+            converter.apply(dense_matrix), matrix_udt)
+        return literal_matrix
+
+
+register_input_converter(OneDimensionalDoubleNumpyArrayConverter(), prepend = True)
```
henrydavidge, line 90: So this runs every time the module is imported? That seems kind of weird. Let's just register once in `glow.register`.

python/glow/conversions.py
```diff
@@ -0,0 +1,86 @@
...
+        return _is_numpy_double_array(object, dimensions = 2)
+
+    def convert(self, object, gateway_client):
+        sc = SparkContext._active_spark_context
+        flat_arr = object.ravel()
+        java_arr = _convert_numpy_to_java_array(flat_arr)
+        dense_matrix = sc._jvm.org.apache.spark.ml.linalg.DenseMatrix(object.shape[0], object.shape[1], java_arr)
+        matrix_udt = sc._jvm.org.apache.spark.ml.linalg.MatrixUDT()
+        converter = sc._jvm.org.apache.spark.sql.catalyst.CatalystTypeConverters.createToCatalystConverter(matrix_udt)
+        literal_matrix = sc._jvm.org.apache.spark.sql.catalyst.expressions.Literal.create(
```
henrydavidge, line 84: Hm, it seems a little weird that we explicitly wrap in a `Literal` here. Does that mean that this conversion doesn't work with `spark.createDataFrame`? Not sure I would expect that as a user.

karenfeng: Unfortunately the PySpark pathways for `createDataFrame` and `lit` are very different. Passing this as a `Literal` allows us to bypass the usual type-checking and conversion logic.

python/glow/conversions.py
```diff
@@ -0,0 +1,104 @@
...
+        return dense_matrix, matrix_udt
+
+# Perform conversions inside literal function calls
+glow_java_converter_list = [ArrayConverter(), MatrixConverter()]
+def lit_decorator(lit_fn):
+    def decorate_lit(col):
+        for java_converter in glow_java_converter_list:
+            if java_converter.can_convert(col):
+                jvm = SparkContext._active_spark_context._jvm
+                java_obj, data_type = java_converter.convert(col)
```
henrydavidge, line 95: nit: can you put the the logic about creating a java literal into a function?

python/glow/conversions.py
```diff
@@ -0,0 +1,104 @@
...
+                catalyst_converter = jvm.org.apache.spark.sql.catalyst.CatalystTypeConverters.createToCatalystConverter(
```
henrydavidge, line 96: It might work to call `Literal.create` directly on the Java object without creating a type converter manually.

karenfeng: We can't use `create[T : TypeTag](v: T)` due to type tags being filled in by the compiler. However, I believe we also can't natively use `Literal.create(v, dataType)`, which depends on the limited types offered in `convertToCatalyst ` versus those in `createToCatalystConverter`. I changed this to directly call the constructor for `Literal` to match the type-tagged implementation.

karenfeng:
> Just a comment. It might be possible to skip the catalyst data converter stuff and rely on `Literal.create(java_obj)`. Also, have you verified that this works in DBR?

This works in DBR if the user runs `from glow import *` or `from glow.conversions import *` before running `from pyspark.sql.functions import *`.

PR #217, opened 2020-05-19
Improve GWAS documentation

karenfeng:

Changes the GWAS documentation to read covariates and phenotypes from CSVs in order to better match a realistic workflow and document that the user can pass literal numpy arrays as a result of https://github.com/projectglow/glow/pull/213.

Due to incompatibility between the most recent release of PySpark 3.0 and the current Spark 3.0.1 snapshot that broke Pandas functionality, I also modified the build to create our own pip-installable `.tar.gz` off the latest Spark `branch-3.0`. This required us to create our own `setup.py` file, as the existing one in Spark requires jars

This change necessitated enabling the env var `PYSPARK_ROW_FIELD_SORTING_ENABLED` due to the changes in https://github.com/apache/spark/pull/26496, which modified the row ordering and broke the Python tests when the row elements were unsorted.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


docs/source/tertiary/regression-tests.rst
```diff
@@ -25,34 +32,55 @@ Example
...
 
-  # Generate random phenotypes and an intercept-only covariate matrix
-  n_samples = df.select(fx.size('genotypes')).first()[0]
-  covariates = DenseMatrix(n_samples, 1, np.ones(n_samples))
-  np.random.seed(500)
-  phenotypes = np.random.random(n_samples).tolist()
-  covariates_and_phenotypes = spark.createDataFrame([[covariates, phenotypes]],
-    ['covariates', 'phenotypes'])
+  # Read phenotypes from a CSV file
+  continuous_phenotypes = pd.read_csv(continuous_phenotypes_csv, index_col=0)
```
henrydavidge, line 55: nit: I think we could just call this `phenotypes`. `continuous_phenotypes` might cause some confusion about the meaning of continuous in this context.

PR #222, opened 2020-05-28
Accept non-string args in transformers

henrydavidge:

Previously, all args had to be strings. Now, we JSON encode non-string args. This makes it easier to use transformers like `block_variants_and_samples` or `pipe`.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


core/src/main/scala/io/projectglow/Glow.scala
```diff
@@ -51,15 +56,20 @@ class GlowBase {
    *
    * @return The transformed DataFrame
    */
-  def transform(operationName: String, df: DataFrame, options: Map[String, String]): DataFrame = {
+  def transform(operationName: String, df: DataFrame, options: Map[String, Any]): DataFrame = {
+    //
```
karenfeng, line 60: Typo?

core/src/main/scala/io/projectglow/Glow.scala
```diff
@@ -51,15 +56,20 @@ class GlowBase {
...
+    val stringValuedMap = options.mapValues {
+      case s: String => s
```
karenfeng, line 62: Do we need this?

core/src/test/scala/io/projectglow/GlowSuite.scala
```diff
@@ -72,13 +72,35 @@ class GlowSuite extends GlowBaseTest {
         ("snake_animal", "snake"))
     checkTransform(df)
   }
+
+  test("accept non-string values") {
```
karenfeng, line 76: I think we could make this test more effective by having the transformer output something that demonstrates the number was parsed properly by the mapper. For example, by checking that the same input parameter passed as a string or as a float results in the same output DF.

core/src/test/scala/io/projectglow/transformers/pipe/PipeTransformerSuite.scala
```diff
@@ -101,6 +101,19 @@ class PipeTransformerSuite extends GlowBaseTest {
...
   }
+
+  test("pass command as a Seq") {
+    val sess = spark
+    import sess.implicits._
+    val inputDf = spark.createDataFrame(Seq(Tuple1("monkey")))
+    val options = Map(
+      "inputFormatter" -> "text",
+      "outputFormatter" -> "text",
+      "cmd" -> Seq("sed", "-e", "s/$/!/"))
```
karenfeng, line 112: Nice! This should makes usage much easier.

core/src/main/scala/io/projectglow/Glow.scala
henrydavidge: Yes, otherwise strings are formatted like `"string"`

core/src/test/scala/io/projectglow/GlowSuite.scala
henrydavidge: Isn't that what the following test does? It verifies that an argument passed as a float can be correctly interpreted. I added a line that checks that a string works the same way.

PR #223, opened 2020-06-01
Fix IntelliJ import

karenfeng:

The existing cross-build functionality does not work well with IntelliJ. To fix this and simplify the build, I changed our cross-build setup to set the Scala version via an environment variable `SCALA_VERSION`. The user will have to change their [SBT settings](https://www.jetbrains.com/help/idea/sbt-support.html) according to the README.

As a result, the `core_2_11` and `core_2_12` projects no longer exist; the Scala base project is accessed via `core` with the appropriate env var. To make this compatible with the release process, we toggle the `scalaVersion` SBT settingKey before running the full suites of tests.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests


build.sbt
```diff
@@ -308,29 +300,37 @@ lazy val stagedRelease = (project in file("core/src/test"))
...
-
-lazy val stagedRelease_2_11 = stagedRelease(scala211)
-lazy val stagedRelease_2_12 = stagedRelease(scala212)
 
 import ReleaseTransformations._
 
 // Don't use sbt-release's cross facility	
 releaseCrossBuild := false
 
+def wrapReleaseStep(step: ReleaseStep): Seq[ReleaseStep] = {
```
henrydavidge, line 309: nit: maybe call it `crossReleaseStep` or something like that?

build.sbt
```diff
@@ -18,7 +19,7 @@ def majorMinorVersion(version: String): String = {
   }
 }
 
-ThisBuild / scalaVersion := scala211
+ThisBuild / scalaVersion := sys.env.getOrElse("SCALA_VERSION", scala211)
```
henrydavidge, line 22: Can we make 212 the default actually? It compiles faster.

PR #240, opened 2020-06-25
Perform GloWGR input validation

karenfeng:

Adds the following input validation steps to GloWGR.

- During the creation of the block genotype matrix, filters out any variants with uniform values. We issue a warning if any rows are dropped. This serves as a preventative measure to the current error that is raised if `sig == 0` during `assemble_block`.
- During any of the ridge model functions:
   + Raise an error if any values are missing in the covariate or label DataFrames
   + Issue a warning if the labels are not mean-centered at 0 with unit variance.

To cut down slightly on runtime, I also wrote out the level 1-reduced GT matrix to use in tests that don't require the original block GT matrix.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


python/glow/wgr/linear_model/functions.py
```diff
@@ -261,3 +264,52 @@ def generate_alphas(blockdf: DataFrame) -> Dict[str, Float]:
...
+    missing values, and should be standardized to zero mean and unit standard deviation. The covariates DataFrame
+    cannot have missing values.
+
+    Args:
+        labeldf : Pandas DataFrame containing target labels
+        covdf : Pandas DataFrame containing covariates
+    """
+    __assert_all_present(labeldf, 'label')
+    __check_standardized(labeldf, 'label')
+    __assert_all_present(covdf, 'covariate')
```
LelandBarnard, line 315: We should also check standardization of covariates here.

karenfeng: Done, thanks for the review! Can you take a second look?

core/src/main/scala/io/projectglow/transformers/blockvariantsandsamples/VariantSampleBlockMaker.scala
```diff
@@ -27,6 +27,16 @@ import org.apache.spark.sql.types.{ArrayType, IntegerType, StringType}
 
 private[projectglow] object VariantSampleBlockMaker extends GlowLogging {
 
+  def filterStdDevZero(df: DataFrame): DataFrame = {
+    val filteredDf = df.filter(size(array_distinct(col(valuesField.name))) > 1)
+    val numOrig = df.count()
```
henrydavidge, line 32: If we're starting with a VCF or Plink file here, this could be expensive because we'd have to parse the input file 3 times (once to count original, once to count filtered, and once for the downstream computation). I'm not sure it's worth it for this warning.

docs/source/tertiary/whole-genome-regression.rst
```diff
@@ -76,21 +75,23 @@ Example
 
     import pandas as pd
 
-    label_df = pd.read_csv(continuous_phenotypes_csv, index_col='sample_id') \
-        .apply(lambda x: x-x.mean())[['Continuous_Trait_1', 'Continuous_Trait_2']]
+    label_df = pd.read_csv(continuous_phenotypes_csv, index_col='sample_id')
+    label_df = ((label_df - label_df.mean())/label_df.std(ddof=0))[['Continuous_Trait_1', 'Continuous_Trait_2']]
```
henrydavidge, line 79: This is fine for now, but we might want to recommend that people just use [sciki-learn's standard scaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html).

core/src/main/scala/io/projectglow/transformers/blockvariantsandsamples/VariantSampleBlockMaker.scala
karenfeng: I considered caching, but I wasn't sure if it was worth it. We can always print a warning; it shouldn't be too hard for users to figure out if they lost any variants.

docs/source/tertiary/whole-genome-regression.rst
karenfeng: Sure - I just didn't want to add an unnecessary dependency just for documentation purposes.

core/src/main/scala/io/projectglow/transformers/blockvariantsandsamples/VariantSampleBlockMaker.scala
```diff
@@ -27,6 +27,11 @@ import org.apache.spark.sql.types.{ArrayType, IntegerType, StringType}
 
 private[projectglow] object VariantSampleBlockMaker extends GlowLogging {
 
+  def filterOneDistinctValue(df: DataFrame): DataFrame = {
+    logger.warn("Filtered variants whose values were all the same.")
```
henrydavidge, line 31: nit: I'd saying "Filtering" to be clear that we didn't necessarily filter anything. This should probably also be an INFO instead of a WARN.

PR #241, opened 2020-06-25
Filter warning for applyInPandas

henrydavidge:

Explicitly ignore this warning. Once we only support Spark 3.x, we should actually heed the warning and use `applyInPandas` instead.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual tests

Will check the unit test logs to confirm the filtering works.


python/glow/wgr/linear_model/ridge_model.py
```diff
@@ -22,6 +22,11 @@
 from typeguard import typechecked
 from typing import Any, Dict, List
 from glow.logging import record_hls_event
+import warnings
+
+# Ignore warning to use applyInPandas instead of apply
+# TODO(hhd): Remove this and start using applyInPnadas once we only support Spark 3.x.
```
karenfeng, line 28: Nit: `applyInPnadas` -> `applyInPandas`

PR #242, opened 2020-06-29
Create LOCO transformation function for RidgeRegression

karenfeng:

Creates a LOCO transformation function for RidgeRegression. This is similar to the existing `transform()`, but with the addition of an optional `chromosomes` parameter. If the parameter is not provided, we extract the chromosomes from the headers in the block DataFrame.

This PR also changes the re-header logic during the RidgeReduction step.
- We no longer include `all`-like terms starting from the 2nd round of reduction, and instead exclude any terms that are no longer relevant.
- We now hash the chromosome to a numeric during the 2nd level of reduction to set the inner index when determining sort keys. This prevents us from breaking on sex chromosomes and any non-numeric chromosome names.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


docs/source/tertiary/whole-genome-regression.rst
```diff
@@ -344,51 +344,41 @@ Model transformation
 ====================
 
 After fitting the ``RidgeRegression`` model, the model DataFrame and cross validation DataFrame are used to apply the
-model to the block matrix DataFrame to produce predictions (*y_hat*) for each label and sample using the
-``RidgeRegression.transform`` method.
+model to the block matrix DataFrame to produce predictions (*y_hat*) for each label and sample based on the
+leave-one-chromosome-out (LOCO) scheme using the ``RidgeRegression.transform_loco`` method.
```
henrydavidge, line 348: nit: I think you could say that you use the fit model df to generate predictions using `transform` or `transform_loco`.

python/glow/wgr/linear_model/ridge_model.py
```diff
@@ -312,6 +312,56 @@ def transform(self,
...
+        Args:
+            blockdf : Spark DataFrame representing the beginning block matrix X
+            labeldf : Pandas DataFrame containing the target labels used in fitting the ridge models
+            sample_blocks : Dict containing a mapping of sample_block ID to a list of corresponding sample IDs
+            modeldf : Spark DataFrame produced by the RidgeRegression fit method, representing the reducer model
+            cvdf : Spark DataFrame produced by the RidgeRegression fit method, containing the results of the cross
+            validation routine.
+            covdf : Pandas DataFrame containing covariates to be included in every model in the stacking
+            ensemble (optional).
+            chromosomes : List of chromosomes to leave out during the LOCO scheme (optional). If not provided, the
```
henrydavidge, line 336: It might be more clear to state this as the chromosomes for which we should generate a prediction.

python/glow/wgr/linear_model/ridge_model.py
```diff
@@ -312,6 +312,56 @@ def transform(self,
...
+            chromosomes will be inferred from the block matrix.
+
+        Returns:
+            Pandas DataFrame containing prediction y_hat values per chromosome. The rows are indexed by sample ID and
+            chromosome; the columns are indexed by label. The column types are float64. The DataFrame is sorted using
+            chromosome as the primary sort key, and sample ID as the secondary sort key.
+        """
+        # Regex captures the chromosome name in the header
+        # level 1 header: chr_3_block_8_alpha_0_label_sim100
+        # level 2 header: all_block_3_alpha_0_label_sim100
```
henrydavidge, line 346: If the first element is `all`, does that not mean "all chromosomes"? It's pretty weird that the chromosome does not come first in that case.

python/glow/wgr/linear_model/ridge_model.py
```diff
@@ -312,6 +312,56 @@ def transform(self,
...
+                       blockdf: DataFrame,
+                       labeldf: pd.DataFrame,
+                       sample_blocks: Dict[str, List[str]],
+                       modeldf: DataFrame,
+                       cvdf: DataFrame,
+                       covdf: pd.DataFrame = pd.DataFrame({}),
+                       chromosomes: List[str] = []) -> pd.DataFrame:
+        """
+        Generates predictions for the target labels in the provided label DataFrame by applying the model resulting from
+        the RidgeRegression fit method to the starting block matrix using a leave-one-chromosome-out (LOCO) scheme.
```
henrydavidge, line 325: nit: I feel like a read "approach" or "method" more commonly than scheme.

python/glow/wgr/linear_model/ridge_model.py
karenfeng: The first term represents the header block, which is `chr_N` for the level-1 reduced block DF and `all` for the level-2 reduced block DF.

karenfeng: @LelandBarnard Would it make sense to replace `all` with `chr_N` for the level-2 reduced block DF?

python/glow/wgr/linear_model/functions.py
```diff
@@ -192,23 +193,31 @@ def new_headers(header_block: str, alpha_names: Iterable[str],
         sort_keys : Array of sortable integers to specify the ordering of the new matrix headers.
         headers : List of new matrix headers.
     """
-    tokens = header_block.split('_')
-
-    if len(tokens) == 2:
-        inner_index = tokens[1]
```
LelandBarnard, line 198: Thank you for doing this!  I hated this function.

python/glow/wgr/linear_model/ridge_model.py
LelandBarnard: Yes, this is a bit confusing.  I think after 2 levels what we really want are headers like `chr_3_alpha_0_label_sim100` or possibly `all_chr_3_alpha_0_label_sim100` but with the hacky way I had written the `new_headers` function it was expecting a pattern like `<outer_block>_<inner_block>_<alpha>_<label>`.  After one round of reduction, the outer block is the chromosome and the inner block is the header block, but after two rounds the outer block is the whole genome (hence the `all`) and the inner block is the chromosome (so `chr_3` becomes `block_3`).  Definitely open to suggestions for cleaning this up.  I think Karen has made a lot of progress already with the improvements to the `new_headers` function.

python/glow/wgr/linear_model/ridge_model.py
```diff
@@ -312,6 +312,55 @@ def transform(self,
...
+        # Regex captures the chromosome name in the header
+        # level 1 header: chr_3_block_8_alpha_0_label_sim100
+        # level 2 header: chr_3_block_all_alpha_0_label_sim100
+        loco_chromosomes = chromosomes if chromosomes else [
+            r.chromosome for r in blockdf.select(
+                f.regexp_extract('header', r"^chr_([a-zA-Z0-9]+)_block", 1).alias(
+                    'chromosome')).distinct().collect()
+        ]
+        loco_chromosomes.sort()
+        print(f'Transforming with a LOCO approach against {loco_chromosomes}')
```
henrydavidge, line 353: Do we print a message for the other functions? It's a bit weird to do it only here. I do think we should add more log messages, but let's handle that separately.

karenfeng: I've been adding logging to functions that infer missing parameters (eg. no `alpha`s) - I'll move this into the inference-specific section.

python/glow/wgr/linear_model/ridge_model.py
karenfeng: Thanks - I took another hack at this to match your first suggestion. Could you take another look? I also hashed the chromosome term if it sets the inner index, so we can match non-numeric chromosome names (eg. X, Y).

PR #243, opened 2020-07-01
Provide function to pivot label DataFrame for gwas functions

henrydavidge:

It's a little bit tricky to pivot labels (is this a better name than phenotypes in general?) into the format that Glow expects for is gwas functions. This PR provides a function to do this pivoting for both the loco and non-loco cases. I'll open a separate PR that updates the docs and our gwas notebooks.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


python/glow/wgr/tests/test_pivot_for_gwas.py
```diff
@@ -0,0 +1,47 @@
...
+        0.2696910984588018, -0.5467797324110817, 1.6258456800754162, 0.03999127669745301,
+        1.093584890813106, -0.5203782912574525, -0.6729752808261074, -0.3071464501504048,
+        -0.016041461027687984, 0.3213354383996257
+    ]
+    assert np.allclose(expected, pt_list[:10])
+
+
+def test_pivot_no_loco(spark):
+    df = pd.read_csv('test-data/wgr/ridge-regression/pts.csv')
+    df = df.set_index('sample_id')
```
karenfeng, line 36: Instead of re-setting `df` here, can you `set_index` immediately following the `read_csv` or use `set_index(inplace=True)`?

python/glow/wgr/tests/test_pivot_for_gwas.py
```diff
@@ -0,0 +1,47 @@
...
+def test_error_too_many_levels(spark):
+    df = pd.DataFrame(columns=['c1', 'c2', 'c3'])
+    df = df.set_index(['c1', 'c2', 'c3'])
+    with pytest.raises(ValueError):
+        functions.pivot_for_gwas(spark, df)
+
+
+def test_pivot_loco(spark):
+    df = pd.read_csv('test-data/wgr/ridge-regression/level1YHatLoco.csv')
+    df = df.set_index(['sample_id', 'contigName'])
```
karenfeng, line 20: Instead of re-setting `df` here, can you `set_index` immediately following the `read_csv` or use `set_index(inplace=True)`? Also, we likely expect the sample IDs and contig names to be strings, which can be done as follows:
```
level1_yhat_loco_df = pd.read_csv(f'{data_root}/level1YHatLoco.csv',
                                  dtype={
                                      'sample_id': 'str',
                                      'contigName': 'str'
                                  }).set_index(['sample_id', 'contigName'])
```
We may want to add a check for this in the pivoting function.

python/glow/wgr/tests/test_pivot_for_gwas.py
```diff
@@ -0,0 +1,47 @@
...
+    pivoted = functions.pivot_for_gwas(spark, df)
+    assert pivoted.count() == 12
+    assert pivoted.columns == ['label', 'contigName', 'values']
+    pt_list = pivoted.where('label = "sim100" and contigName = 1').head().values
+    assert len(pt_list) == 100
+    expected = [
```
karenfeng, line 26: Can you add a check that the list ordering matches the input sample ID order? This would also allow us to avoid hard-coding these values.

python/glow/wgr/functions.py
```diff
@@ -128,3 +129,38 @@ def block_variants_and_samples(variant_df: DataFrame, sample_ids: List[str],
     output = blocked_gt, index_map
     assert check_return_type(output)
     return output
+
+
+def pivot_for_gwas(spark: SparkSession, label_df: pd.DataFrame) -> DataFrame:
+    """
+    Pivots a Pandas DataFrame into a Spark DataFrame with a convenient format for Glow's GWAS
```
karenfeng, line 136: I'm a little iffy about the term "pivot" here; generally, a pivot table contains summary statistics. I know Pandas uses `pivot` when reshaping though.

python/glow/wgr/tests/test_pivot_for_gwas.py
henrydavidge: I actually don't think that we need to or should require this. If someone wants to treat contig names as ints, for instance, that should be totally fine.

henrydavidge:
@karenfeng Comments addressed. Thanks for the testing suggestion -- that actually caught a bug 🤦 . Also, I changed the terminology to "reshape". Let me know if you like that better.

python/glow/wgr/functions.py
```diff
@@ -157,7 +158,8 @@ def pivot_for_gwas(spark: SparkSession, label_df: pd.DataFrame) -> DataFrame:
         transposed_df = label_df.T
         column_names = ['label', 'values']
     elif label_df.index.nlevels == 2:  # Indexed by sample id and contig name
-        transposed_df = label_df.T.stack()
+        sample_ordering = pd.unique(label_df.index.get_level_values(0))
+        transposed_df = label_df.T.stack().reindex(sample_ordering, axis=1)
```
karenfeng, line 162: What happens if we use `.reindex(columns=labeldf.index)` here?

python/glow/wgr/functions.py
```diff
@@ -157,7 +158,8 @@ def pivot_for_gwas(spark: SparkSession, label_df: pd.DataFrame) -> DataFrame:
         transposed_df = label_df.T
```
karenfeng, line 158: Does this guarantee ordering?

python/glow/wgr/functions.py
henrydavidge: If the original index was a multi-index on sample id and contig name, so unfortunately that does not work. We need to the `unique` because the sample ids appear for each chromosome.

python/glow/wgr/functions.py
henrydavidge: Yes, transposing does not sort the columns or rows.

python/glow/wgr/functions.py
```diff
@@ -158,8 +158,10 @@ def reshape_for_gwas(spark: SparkSession, label_df: pd.DataFrame) -> DataFrame:
...
         column_names = ['label', 'values']
     elif label_df.index.nlevels == 2:  # Indexed by sample id and contig name
-        sample_ordering = pd.unique(label_df.index.get_level_values(0))
-        transposed_df = label_df.T.stack().reindex(sample_ordering, axis=1)
+        # stacking orders by the new index, so we need to remember the original sample order in case
```
karenfeng, line 161: I'm a little confused by the grammar here - can you rephrase?

henrydavidge: Done

PR #247, opened 2020-07-07
Update GlowGR notebook and gwas docs

henrydavidge:

- Show Python API docs for GloWGR classes and functions
- Update GloWGR notebook to include new helper functions like `transform_loco` and `reshape_for_gwas`
- Update gwas docs to use `reshape_for_gwas`

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

(Details)


docs/source/api-docs/glowgr-functions.rst
```diff
@@ -0,0 +1,8 @@
+GloWGR
```
karenfeng, line 1: The other high-level function API doc pages have titles ending with `Functions`, should we match that here?

docs/source/api-docs/glowgr-functions.rst
```diff
@@ -0,0 +1,8 @@
...
+------
+
+.. automodule:: glow.wgr.linear_model.ridge_model
```
karenfeng, line 4: I think the ordering might make more sense if we put the general functions on top, followed by the linear-model specific ones below. It might also help to have a little exposition for each class.

docs/source/api-docs/glowgr-functions.rst
henrydavidge: Since we have classes here as well, I don't think we should have `Functions`

docs/source/api-docs/glowgr-functions.rst
henrydavidge: Did you take a look at the rendered documentation? We include the docstring, which is pretty thorough.

henrydavidge: If more exposition is required, it should go in the docstring.

karenfeng: I was thinking more in terms of explicitly denoting the different subsections - the first section is for general use, the second section is for the linear model, and the last section will be for the binary model. But we can clean that up when the binary model goes in.

henrydavidge: Ah, so just like section headers? I'll add that.

PR #251, opened 2020-07-13
update arch diagram

williambrandler:

Added gff to architecture diagram in intro

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


williambrandler:
@kianfar77 could you take a look at this architecture diagram and let me know what you think? Cheers!

kianfar77:
> @kianfar77 could you take a look at this architecture diagram and let me know what you think? Cheers!

Thanks William for updating this. I have a few suggestions: 

- I would put the entry to the Bronze level in two categories: Variants and Annotations. For Variants, we can put VCF, BGEN, And PLINK  and for Annotations we can put GFF3.
- If so then the liftover and normalization can be put on the line going into bronze from the variant category. 
- I would also add "Splitting multiallelic variants" somewhere as we have a pretty powerful parallelized splitter in glow, perhaps with liftover and normalization or may be as something done before GWAS. 
- I think it is better to have liftover then normalize just to be sure that normalization happens on the final reference genome.
- I would add GloWGR to GWAS as well. 
- Minor: subset -> Subset (for consistency); Table -> Tables (for Bronze and Silver as it can be more than one delta table)

williambrandler:
cheers! Updated, let me know what you think

williambrandler:
thanks @kianfar77, I added in GWAS, the Spark logo, and a tagline

karenfeng:
Merging! Thanks @williambrandler.

PR #264, opened 2020-07-27
Non-compressed VCFs should be splitable

henrydavidge:

We missed this bit when we forked `TextBasedFileFormat`'s splitability conditions.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
+++ b/core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
@@ -67,7 +67,8 @@ class VCFFileFormat extends TextBasedFileFormat with DataSourceRegister with Hls
 
     // Note: we check if a file is gzipped vs block gzipped during reading, so this will be true
     // for .gz files even if they aren't actually splittable
-    codecFactory.getCodec(path).isInstanceOf[SplittableCompressionCodec]
+    val codec = codecFactory.getCodec(path)
+    codec == null || codecFactory.getCodec(path).isInstanceOf[SplittableCompressionCodec]
   }
 
   override def inferSchema(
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
```diff
@@ -67,7 +67,8 @@ class VCFFileFormat extends TextBasedFileFormat with DataSourceRegister with Hls
 
     // Note: we check if a file is gzipped vs block gzipped during reading, so this will be true
     // for .gz files even if they aren't actually splittable
-    codecFactory.getCodec(path).isInstanceOf[SplittableCompressionCodec]
+    val codec = codecFactory.getCodec(path)
+    codec == null || codecFactory.getCodec(path).isInstanceOf[SplittableCompressionCodec]
```
karenfeng, line 71: Is this meant to be the same logic as https://github.com/apache/spark/blob/94fc0e3235162afc6038019eed6ec546e3d1983e/core/src/main/scala/org/apache/spark/util/Utils.scala#L2921?
What is the value of `codec` in the case that different extensions are passed in? It looks like the value could be `null` if the extension is unrecognized: https://hadoop.apache.org/docs/r2.4.1/api/src-html/org/apache/hadoop/io/compress/CompressionCodecFactory.html#line.192

henrydavidge: It's the same logic as `TextBasedFileFormat`. To reduce duplication, I changed my code to explicitly call `super.isSplittable`. I think Spark treats unrecognized extensions as plain text. There are many text based file formats that can be interpreted as plain text. It's not really possible to enumerate all of them.

henrydavidge:
Merging since only the link checker failed and @karenfeng has fixed that.

Issue #255, opened 2020-07-20
WGR small inconsistencies and possible improvements


eric-czech:
Hi, 

I was looking through the preprint and some of the code for the new WGR functionality and I love the work you guys have done on it -- thank you for sharing that all!  The work that went into the documentation is much appreciated too, especially the [plain-text descriptions](https://glow.readthedocs.io/en/latest/tertiary/whole-genome-regression.html) (an all too uncommon bridge between methods of a paper and the code itself IMO).

I had a few thoughts/suggestions that I thought I would share in looking at the code:

- **Matrix Inversion**
  - Have you considered using something other than LU decomposition (via np.linalg.inv) for inverting (XtX + λI)?  I worry that even though the whole genotype matrix would rarely be singular, I think it would be reasonable to expect that certain blocks of it may be
    - I think one good solution could be to swap `inv` for `pinv` to use an SVD solver instead in: https://github.com/projectglow/glow/blob/f3edf5bb8fe9c2d2e1a374d4402032ba5ce08e29/python/glow/wgr/linear_model/functions.py#L154
    - Some experiments on small datasets produced the exact same results this way when all blocks are non-singular
    - Another could be to use the fact that (XtX + λI) is symmetric and switch to Cholesky decomposition 
    - I'm not very familiar with the performance characteristics of each, but I know the computational complexity of pseudo-inverses and cholesky can be lower than LU and even if they don't make for a big improvement in speed, I think the stability and robustness to a single block with perfectly correlated columns would be worth it
- **Genotype Scaling**
  - I saw that the [demo notebook](https://glow.readthedocs.io/en/latest/_static/notebooks/tertiary/glowgr.html) standardizes on 0 degrees of freedom for `std` calculations while the stddev calculated for the genotypes matrix itself does not: https://github.com/projectglow/glow/blob/75ea111e0ba86001ea213df103373125b8732778/core/src/main/scala/io/projectglow/sql/expressions/MomentAggState.scala#L68
    - It's a very minor point, but perhaps this should be `count` instead of `count - 1` to be consistent
- **Reduced Block Scaling**
  - It appears that the stage 1 predictions are standardized on a per-sample-block basis rather than across all samples as stated in the preprint and in the c++ code at: https://github.com/rgcgithub/regenie/blob/master/src/Step1_Models.cpp#L250
[...]

PR #268, opened 2020-08-03
Make genotype scaling consistent

karenfeng:

Addresses #255 by making genotype statistics use population standard deviation instead of sample standard deviation, as we use the former for both phenotype and covariate scaling.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests


karenfeng:
Closing as it would be preferable to scale to sample standard deviation.

PR #274, opened 2020-08-07
Faster VCF reader

henrydavidge:
This PR introduces a Glow-specific VCF reader that converts VCF lines directly into `InternalRow`s without going through htsjdk. In addition to much improved performance (benchmarks to follow), we can now flexibly handle partially missing DP and PL values. Htsjdk nulls out the entire array if any values are missing.

I did introduce a behavior change regarding empty arrays. Previously, in some cases we'd treat array fields with a single `.` as `null` and other times as an empty array. Now, we always return `null`. This change is not necessary and can be reverted.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

All unit tests pass with the new VCF reader enabled. I also verified that it works with the Databricks joint genotyping pipeline works with the new VCF reader.


core/src/main/scala/io/projectglow/vcf/VCFLineToInternalRowConverter.scala
```diff
@@ -0,0 +1,517 @@
...
+import org.apache.spark.unsafe.types.UTF8String
+import org.apache.spark.unsafe.types.UTF8String.{IntWrapper, LongWrapper}
+import io.projectglow.common.{GenotypeFields, HasStringency, SimpleInterval, VariantSchemas}
+
+import scala.util.control.NonFatal
+
+/**
+ * Converts the raw bytes in a VCF line into an [[InternalRow]]
+ *
+ * @param header The VCF header object this is currently only used to extract sample ideas
```
kianfar77, line 41: ideas -> ids

core/src/main/scala/io/projectglow/vcf/VCFFileFormat.scala
```diff
@@ -150,6 +152,19 @@ class VCFFileFormat extends TextBasedFileFormat with DataSourceRegister with Hls
...
+    val hasAttributesField =
+      requiredSchema.exists(
+        SQLUtils.structFieldsEqualExceptNullability(_, VariantSchemas.attributesField))
+    val hasOtherFields =
+      InternalRowToVariantContextConverter.getGenotypeSchema(requiredSchema) match {
+        case None => false
+        case Some(schema) =>
+          schema.exists(
+            SQLUtils.structFieldsEqualExceptNullability(_, VariantSchemas.otherFieldsField))
+      }
```
kianfar77, line 164: Can we add a comment about the current limitations of the fast reader here?

core/src/main/scala/io/projectglow/vcf/VCFLineToInternalRowConverter.scala
```diff
@@ -0,0 +1,517 @@
...
+      return row
+    }
+
+    ctx.expectTab()
+    row.update(genotypesIdx, parseGenotypes(ctx))
+
+    row
+  }
+
+  private def parseGenotypes(ctx: LineCtx): GenericArrayData = {
```
kianfar77, line 208: Can we have a couple of lines documenting the behavior of this parser?

kianfar77 (commented):
Looks great. A few minor comments and a question.

core/src/test/scala/io/projectglow/vcf/VCFDatasourceSuite.scala
```diff
@@ -178,121 +176,161 @@ class VCFDatasourceSuite extends GlowBaseTest {
...
+      .selectExpr("genotypes[0].calls")
+      .as[Seq[Int]]
+      .head
+    assert(calls == Seq(-1))
   }
 
   test("split uncalled genotype") {
-    val row = parseVcfContents(makeVcfRow(Seq("AC=2", "GT", "./."))).head
-    assert(row.genotypes.head.calls.get == Seq(-1, -1))
+    import sess.implicits._
```
kianfar77, line 219: Can we change the term "split" in the name of this test?

core/src/test/scala/io/projectglow/vcf/VCFDatasourceSuite.scala
```diff
@@ -85,7 +85,6 @@ class VCFDatasourceSuite extends GlowBaseTest {
   }
 
   test("check parsed row") {
```
kianfar77, line 87: I think we should have something like this for FastReader too where the assertion is against row with flattened info fields.

core/src/test/scala/io/projectglow/vcf/VCFDatasourceSuite.scala
```diff
@@ -693,6 +738,33 @@ class VCFDatasourceSuite extends GlowBaseTest {
...
+
+  test("prune genotype fields") {
+    import sess.implicits._
+    val extraHeaderLines =
+      """
+        |##FORMAT=<ID=MIN_DP,Number=1,Type=Integer,Description="">
+        |##FORMAT=<ID=AD,Number=R,Type=Integer,Description="">
+        |""".stripMargin.trim() + "\n"
+    val df = parseVcfContents(
+      "chr0\t0\t.\tA\t.\t.\t.\t.\tAD:MIN_DP\t2,5:1",
```
kianfar77, line 750: Nor immediately clear to me why `makeVcfLine` is not used like other tests and instead the first portion of the file  is provided directly.

core/src/test/scala/io/projectglow/vcf/VCFDatasourceSuite.scala
```diff
@@ -693,6 +738,33 @@ class VCFDatasourceSuite extends GlowBaseTest {
...
+}
+
+class FastVCFDatasourceSuite extends VCFDatasourceSuite {
+  override def sparkConf: SparkConf =
+    super.sparkConf.set(GlowConf.FAST_VCF_READER_ENABLED.key, "true")
+
+  test("read AD with nulls") {
+    import sess.implicits._
+    val df = parseVcfContents(
+      "chr0\t0\t.\tA\t.\t.\t.\t.\tAD\t.,1",
```
kianfar77, line 763: same comment as above

core/src/main/scala/io/projectglow/vcf/VCFLineToInternalRowConverter.scala
```diff
@@ -0,0 +1,532 @@
...
+/**
+ * Converts the raw bytes in a VCF line into an [[InternalRow]]
+ *
+ * @param header The VCF header object this is currently only used to extract sample ids
+ * @param schema The schema of the converted rows
+ * @param stringency How to handle errors
+ * @param overlapDetectorOpt If provided, the converter will check to see if a variant passes this detector before
+ *                           parsing genotypes
+ */
+class VCFLineToInternalRowConverter(
```
kianfar77, line 47: If I have not missed it, it seems validations against the `Number` for each field in the header is lost in this reader. I would assume `VariantContext` should have such validation. For example, if `Number = R` for a genotype fied do we check that is actually the case and perhaps warn or even create nulls in the array?

kianfar77: Never mind this. I checked and it is coded in VariantContext but is not used.

core/src/test/scala/io/projectglow/vcf/VCFDatasourceSuite.scala
henrydavidge: I adapted the suite that checks flattened info fields to check the entire parsed row.

core/src/test/scala/io/projectglow/vcf/VCFDatasourceSuite.scala
```diff
@@ -395,25 +395,47 @@ class VCFDatasourceSuite extends GlowBaseTest {
...
+          Some("NA12878"),
+          Some(false),
+          Some(List(1, 1)),
+          Some(84),
+          Some(List(3813, 256, 0)),
+          Some(99),
+          Some(List(0, 84))))
+    )
+
+    assert(vc.copy(qual = None) == expected.copy(qual = None))
```
kianfar77, line 437: Can you add a comment on why qual is not precise?

kianfar77 (approved):
Great PR! Approved subject to addition of the minor comment.

PR #276, opened 2020-08-10
Support reading BGEN files with uncompressed or zstd compressed genotypes

henrydavidge:
Some newer tools now write zstd compressed files by default.

cc @bboutkov 

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


bboutkov:
Cool! Thanks, and I appreciate the extra support for uncompressed BGENs - it's nice to be able to avoid rounding issues despite however small the error may be. 

Not sure about the missing header issue - Im guessing some dependency versioning issue? Once that's cleared up looks good to me.

core/src/main/scala/io/projectglow/bgen/BgenGenotypeReader.scala
```diff
@@ -0,0 +1,55 @@
...
+import java.util.zip.Inflater
+
+import com.github.luben.zstd.{Zstd => ZstdCodec}
+
+trait BgenGenotypeReader {
+  /**
+   * Read genotypes from a bgen [[DataInput]] into a byte array. Returns the uncompressed genotype bytes.
+   * Note that the genotypes returned are still encoded using the standard BGEN encoding scheme.
+   */
+  def readGenotypes(dataStream: DataInput): Array[Byte]
```
karenfeng, line 13: Can we rename this to `readGenotypeBlock`? We should also be clear this is only for Layout 2 blocks.

core/src/main/scala/io/projectglow/bgen/BgenGenotypeReader.scala
```diff
@@ -0,0 +1,55 @@
+package io.projectglow.bgen
```
karenfeng, line 1: I think the tests are failing because we're missing the license header.

Issue #280, opened 2020-08-13
Document disabling whole-stage codegen during GloWGR data prep


karenfeng:
@henrydavidge noticed that if multiallelic splitting is enabled during the block matrix creation, the generated method exceeds the JVM JIT limit. This results in a fall back to interpreted mode, which is much slower. We should document disabling whole-stage codegen via `spark.conf.set("spark.sql.codegen.wholeStage", codegen)`.

PR #282, opened 2020-08-17
GloWGR scaling improvements

karenfeng:

- Pushes validation that each row of variants contains the same number of genotype values lazily into the block transformer.
- Disable automatic broadcast merges that are likely to OOM, such as`Caused by: org.apache.spark.sql.execution.OutOfMemorySparkException: Size of broadcasted table far exceeds estimates and exceeds limit of spark.driver.maxResultSize=4294967296. You can disable broadcasts for this query using set spark.sql.autoBroadcastJoinThreshold=-1`
- Adds logging during `transform_loco` for the per-chromosome predictions.
- Adds documentation advising against performing multiallelic splitting in the same query as matrix blocking, related to #280.
- Adds documentation regarding scalability limits. PyArrow indexes its elements with integers; as a result, the size of the vector buffer is limited to the maximum value of an integer (`2,147,483,647`). Vectors are allocated in sizes that are powers of two; the largest power-of-two below this limit is `1,073,741,824`. Each vector buffer contains a data buffer and a validity buffer. The size of the data buffer of 8-byte floats is `(# elements * 8)` and the size of the validity buffer is `(# elements + 63) >> 6) << 3.` Therefore, the maximum number of elements is `132,152,839`. The bottleneck of elements in each sample block/label pair is determined by the following equation: `(# alphas) * (# SNPs/ # SNPs per block) * (# samples / # sample blocks)`. The floats are stored in an array of length `(# samples / # sample blocks)` in each row. There are `(# SNPs/ # SNPs per block) * (# alphas)` rows.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests


core/src/test/scala/io/projectglow/transformers/blockvariantsandsamples/BlockVariantsAndSamplesTransformerSuite.scala
```diff
@@ -150,4 +153,18 @@ class BlockVariantsAndSamplesTransformerSuite extends GlowBaseTest with GlowLogg
...
+      .read
+      .format("vcf")
+      .load(testVcf)
+      .withColumn("values", expr("array_repeat(0.0, INFO_DP)"))
+    val options = Map(VARIANTS_PER_BLOCK -> "10", SAMPLE_BLOCK_COUNT -> "20")
+    val ex = intercept[SparkException] {
+      Glow.transform(TRANSFORMER_NAME, vcfDf, options).show()
+    }
+    assert(ex.getCause.isInstanceOf[RuntimeException])
+    assert(ex.getCause.getMessage.contains("is not true"))
```
henrydavidge, line 168: What does the error message look like here?

docs/source/tertiary/whole-genome-regression.rst
```diff
@@ -380,6 +382,19 @@ Example
...
+Troubleshooting
+---------------
+
+If you encounter limits related to memory allocation in PyArrow, you may need to tune the number of alphas, number of
+variants per block, and/or the number of sample blocks. The default values for these hyperparameters are tuned for
+500,000 variants and 500,000 samples.
+
+
+The known limit is:
+``(# alphas) * (# variants/ # variants per block) * (# samples / # sample blocks) <= 132,152,839``
```
henrydavidge, line 395: Isn't there another limit based on the total number of header blocks?

python/glow/wgr/linear_model/ridge_model.py
```diff
@@ -281,7 +281,7 @@ def transform(self,
             PandasUDFType.GROUPED_MAP)
 
         blocked_prediction_df = blockdf.drop('header_block', 'sort_key') \
-            .join(modeldf.drop('header_block'), ['sample_block', 'header'], 'right') \
+            .join(modeldf.drop('header_block').hint('merge'), ['sample_block', 'header'], 'right') \
```
henrydavidge, line 284: Add a comment about the `'merge'` plz

docs/source/tertiary/whole-genome-regression.rst
karenfeng: The total number of header blocks meaning `(# variants/ # variants per block) * (# sample blocks)`? I haven't encountered bottlenecks during the ridge regression (versus the reduction) step during my testing. This is the limit for any Arrow vector, so we can check that any assembled group during GloWGR is smaller than this scale.

henrydavidge: I see, actually the expression here does depend on the total number of header blocks as you say. Is that definitely during the reduction step? I thought that during reduction we never load in more than one block of the matrix at a time, so the total number of variants should not matter.

Also, I'm very surprised that the number of labels is not in this expression.

core/src/test/scala/io/projectglow/transformers/blockvariantsandsamples/BlockVariantsAndSamplesTransformerSuite.scala
karenfeng: It looks like this: ```java.lang.RuntimeException: '(size(array_repeat(0.0, input[4, int, true]), true) = 17190)' is not true!```

docs/source/tertiary/whole-genome-regression.rst
karenfeng: Sorry, I wrote my last statement incorrectly - this is a bottleneck I encountered during the ridge regression step. In particular, it's in the map step after grouping by sample block and label (hence why the number of labels doesn't matter): https://github.com/projectglow/glow/blob/50afeadb5d980c7216517e1ce7c8572b467d3d02/python/glow/wgr/linear_model/ridge_model.py#L237

In this case, the number of variant blocks (`# variants/ # variants per block`) does matter.

core/src/test/scala/io/projectglow/transformers/blockvariantsandsamples/BlockVariantsAndSamplesTransformerSuite.scala
henrydavidge: Hm, that's an inscrutable error message. If we give the column an informative name, does the alias appear in the error message?

karenfeng: The column is already called `values`; maybe I can re-throw with a more helpful message.

karenfeng: Actually, I'm not sure if we can re-throw in this case as the issue doesn't manifest until the DataFrame is materialized.

core/src/main/scala/io/projectglow/functions.scala
```diff
@@ -341,6 +341,19 @@ object functions {
...
+  /**
+   * Asserts a boolean condition is true.
+   * @group quality_control
+   * @since 0.5.0
+   *
+   * @param condition Boolean condition to check
+   * @param errMsg Error message if condition fails
+   * @return Null if true, or throws an exception if not true
+   */
+  def assert_true_or_error(condition: Column, errMsg: String): Column = withExpr {
```
henrydavidge, line 353: Hm, actually I think it might be better to exclude this from the Python / Scala APIs. I think the use case is pretty narrow, and since this will be in the next release of Spark, I don't think there's much upside to exposing it in our public API.

You can set the `exclude_python` option in the YAML file. We should add an analogous `exclude_scala` flag. I can take care of that if you don't have time.

core/src/main/scala/io/projectglow/transformers/blockvariantsandsamples/VariantSampleBlockMaker.scala
```diff
@@ -20,13 +20,28 @@ import io.projectglow.common.GlowLogging
...
 
+  def validateNumValues(df: DataFrame): DataFrame = {
+    import df.sparkSession.implicits._
+    val expectedNumValues =
+      df.selectExpr("size(values) as numValues").take(1)(0).getAs[Int]("numValues")
+    df.filter(
+      isnull(
+        assert_true_or_error(
+          size(col("values")) === expectedNumValues,
+          "Number of values is inconsistent!")))
```
henrydavidge, line 42: How about "At least one row has an inconsistent number of values (expected x). Please verify that each row contains the same number of values."

henrydavidge (commented):
Sorry, had a few more comments.

henrydavidge (approved):
Nice! LGTM

PR #283, opened 2020-08-19
Validate array types when converting from variant contexts to internal rows

henrydavidge:
Previously, when converting variant contexts to internal rows, we could throw confusing errors if the inner type of an array typed format field was not consistent between the expected schema and the variant context. Rather than receive a clear error that we expected type x for field y, the user would see a cryptic internal Spark error because of the type mismatch.

Now, we explicitly validate the elements in the variant context's collection type.

Note that compiler warnings could have tipped us off to the fact that we allowed bad values into the internal rows. As a future task, we should make compiler warnings (except probably deprecation warnings) fatal.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
+++ b/core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
@@ -601,7 +601,7 @@ class VariantContextToInternalRowConverter(
     obj match {
       case null => null
       case VCFConstants.MISSING_VALUE_v4 => null
-      case arr: Array[T] =>
+      case arr: Array[AnyRef] =>
         var i = 0
         while (i < arr.length) {
           require(
@@ -611,8 +611,9 @@ class VariantContextToInternalRowConverter(
         }
 
         arr.asInstanceOf[Array[Any]]
-      case arr: Array[R] if primitiveConverter.isDefined => arr.map(primitiveConverter.get)
-      case l: JList[T] =>
+      case arr: Array[_] if primitiveConverter.isDefined =>
+        arr.map(el => primitiveConverter.get(el.asInstanceOf[R]))
+      case l: JList[_] =>
         val arr = new Array[Any](l.size)
         var i = 0
         while (i < arr.length) {
@@ -631,7 +632,7 @@ object VariantContextToInternalRowConverter {
   def parseObjectAsString(obj: Object): String = {
     obj match {
       case d: JDouble => d.toString
-      case dArray: Array[Double] =>
+      case dArray: Array[_] =>
         val sArray = new Array[String](dArray.length)
         var i = 0
         while (i < dArray.length) {
@@ -639,7 +640,7 @@ object VariantContextToInternalRowConverter {
           i += 1
         }
         VCFEncoderUtils.formatVCFField(sArray)
-      case dList: JList[Double] =>
+      case dList: JList[_] =>
         val sArray = new Array[String](dList.size)
         var i = 0
         while (i < dList.size) {
--- a/core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
+++ b/core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
@@ -602,7 +602,7 @@ class VariantContextToInternalRowConverter(
       case arr: Array[T] =>
         var i = 0
         while (i < arr.length) {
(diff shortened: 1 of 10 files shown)


core/src/test/scala/io/projectglow/vcf/VariantContextToInternalRowConverterSuite.scala
```diff
@@ -175,6 +175,33 @@ class VariantContextToInternalRowConverterSuite extends VCFConverterBaseTest {
...
+    def expectFailure(gp: Any): Unit = {
+      val vc = vcb.genotypes(new GenotypeBuilder().attribute("GP", gp).make()).make()
+      intercept[IllegalArgumentException] {
+        strictConverter.convertRow(vc, isSplit = false)
+      }
+    }
+
+    expectFailure(new JArrayList[Integer](Seq(1: Integer).asJavaCollection))
+    expectFailure(Array[Integer](1: Integer))
+    expectFailure(Array[Int](1))
```
karenfeng, line 193: Let's also add a success case as a sanity check, and to clarify why these are failing.

core/src/test/scala/io/projectglow/vcf/VariantContextToInternalRowConverterSuite.scala
```diff
@@ -175,6 +175,33 @@ class VariantContextToInternalRowConverterSuite extends VCFConverterBaseTest {
...
+  }
+
+  test("convert primitive array") {
+    val vcb = new VariantContextBuilder()
+    vcb.chr("").start(1).stop(1)
+    val refAllele = Allele.create("A", true)
+    vcb.alleles(Seq(refAllele).asJava)
+    val vc = vcb.genotypes(new GenotypeBuilder().attribute("GP", Array(1.0d)).make()).make()
+    strictConverter.convertRow(vc, isSplit = false)
```
karenfeng, line 202: Should we check that this did what we expected?

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
```diff
@@ -594,18 +595,32 @@ class VariantContextToInternalRowConverter(
...
+      primitiveConverter: Option[R => T] = None)(implicit ct: ClassTag[T]): Array[Any] =
     obj match {
       case null => null
       case VCFConstants.MISSING_VALUE_v4 => null
-      case arr: Array[T] => arr.asInstanceOf[Array[Any]]
-      case arr: Array[R] if primitiveConverter.isDefined => arr.map(primitiveConverter.get)
-      case l: JList[T] =>
+      case arr: Array[AnyRef] =>
+        var i = 0
+        while (i < arr.length) {
```
karenfeng, line 606: Do we have to check for every element, or is just the head enough?

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
```diff
@@ -594,18 +595,32 @@ class VariantContextToInternalRowConverter(
...
+
+        arr.asInstanceOf[Array[Any]]
+      case arr: Array[_] if primitiveConverter.isDefined =>
+        arr.map(el => primitiveConverter.get(el.asInstanceOf[R]))
+      case l: JList[_] =>
         val arr = new Array[Any](l.size)
         var i = 0
         while (i < arr.length) {
           arr(i) = l.get(i)
+          require(
```
karenfeng, line 621: Same question here - do we have to check every element?

core/src/test/scala/io/projectglow/vcf/VariantContextToInternalRowConverterSuite.scala
henrydavidge: That's why I added the following test

core/src/main/scala/io/projectglow/vcf/VariantContextToInternalRowConverter.scala
henrydavidge: It is possible for each element to have a different type if the original collection type was like `Array[Object]`. Knowing htsjdk, I wouldn't count against that.

karenfeng: OOC, does this add to the runtime significantly?

PR #287, opened 2020-08-26
Use python functions instead of string expressions in docs

henrydavidge:

Some of our docs pages were written before we generated Python functions for our various expressions. Now that we have a more thorough Python API, we should use it in the documentation examples.

Also, I clarified the conversion to a 0-based coordinate system when ingesting VCF files.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

(Details)


henrydavidge:
That makes sense. Let's plan on doing that in a followup.

PR #291, opened 2020-09-04
Spark Cross Release

kianfar77:

This PR modifies build.sbt to cross release glow for 
- Spark3 Scala 2.12 
- Spark2 Scala 2.12
- Spark2 Scala 2.11.

## How is this patch tested?
I ran a round of test release and all worked well. 

(Details)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -323,7 +323,7 @@ lazy val stagedRelease = (project in file("core/src/test"))
     unmanagedSourceDirectories in Test += baseDirectory.value / "shim" / majorMinorVersion(
       sparkVersion.value),
     libraryDependencies ++= testSparkDependencies.value ++ testCoreDependencies.value :+
-    "io.projectglow" %% s"glow_spark${majorVersion(sparkVersion.value)}_scala" % stableVersion.value % "test",
+    "io.projectglow" %% s"glow-spark${majorVersion(sparkVersion.value)}-scala" % stableVersion.value % "test",
     resolvers := Seq("bintray-staging" at "https://dl.bintray.com/projectglow/glow"),
     org
       .jetbrains
--- a/build.sbt
+++ b/build.sbt
@@ -378,8 +378,8 @@ crossReleaseStep(runTest, true) ++
 Seq(
   setReleaseVersion,
   updateStableVersion,
-//    commitReleaseVersion,
-//    commitStableVersion,
+  commitReleaseVersion,
+  commitStableVersion,
   tagRelease
 ) ++
 crossReleaseStep(publishArtifacts, false) ++
--- a/build.sbt
+++ b/build.sbt
@@ -354,7 +354,7 @@ def crossReleaseStep(step: ReleaseStep, requiresPySpark: Boolean): Seq[ReleaseSt
     if (requiresPySpark) "changePySparkVersion" else "")
 
   Seq(
-//    updateCondaEnvStep,
+    updateCondaEnvStep,
 //    releaseStepCommandAndRemaining(s"""set ThisBuild / sparkVersion := "$spark3""""),
 //    releaseStepCommandAndRemaining(s"""set ThisBuild / scalaVersion := "$scala212""""),
 //    step,
--- a/build.sbt
+++ b/build.sbt
@@ -340,7 +340,7 @@ releaseCrossBuild := false
 
 lazy val changePySparkVersion = taskKey[Unit]("changePySparkVersion ")
 changePySparkVersion := {
-  s"conda remove -n glow pyspark ; conda install -n glow pyspark=$sparkOld" !
+  "conda remove -n glow pyspark" !; s"conda install -n glow pyspark=$sparkOld" !
 }
 
 lazy val updateCondaEnv = taskKey[Unit]("Update Glow Env To Latest Version")
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,18 @@ import sbt.nio.Keys._
 lazy val scala212 = "2.12.8"
 lazy val scala211 = "2.11.12"
 
-lazy val sparkVersion = sys.env.getOrElse("SPARK_VERSION", "3.0.0")
+lazy val spark2 = "2.4.5"
+lazy val spark3 = "3.0.0"
+
+lazy val sparkVersion = settingKey[String]("sparkVersion")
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -312,15 +335,36 @@ lazy val stagedRelease = (project in file("core/src/test"))
...
+
+lazy val updateCondaEnv = taskKey[Unit]("Update Glow Env To Latest Version")
+updateCondaEnv := {
+  "conda env update -f python/environment.yml" !
+}
+
+def crossReleaseStep(step: ReleaseStep, test: Boolean): Seq[ReleaseStep] = {
+  val updateCondaEnvStep = releaseStepCommandAndRemaining(if (test) "updateCondaEnv" else "")
+  val downVersionPySparkStep = releaseStepCommandAndRemaining(
+    if (test) "downVersionPySpark" else "")
```
kianfar77, line 354: I assume and my tests verify that pyspark is only needed at the ``runTest`` stage and not at ``publishArtifacts`` or ``stagedRelease/test``. So I set pyspark version changes to be applied only for ``runTest``. Let me know if this is not true.

henrydavidge:
@kianfar77 I think we're missing the script to change the pyspark version?

build.sbt
```diff
@@ -312,15 +335,36 @@ lazy val stagedRelease = (project in file("core/src/test"))
...
+downVersionPySpark := {
+  "./downversion-pyspark.sh" !
+}
+
+lazy val updateCondaEnv = taskKey[Unit]("Update Glow Env To Latest Version")
+updateCondaEnv := {
+  "conda env update -f python/environment.yml" !
+}
+
+def crossReleaseStep(step: ReleaseStep, test: Boolean): Seq[ReleaseStep] = {
```
henrydavidge, line 351: Might be more clear to call this parameter "requiresPyspark" or "updatePysparkVersion"

henrydavidge (commented):
Looks reasonable to me. I'll give @karenfeng a chance to take a look if she has time. If not, I'll give the final approval.

build.sbt
```diff
@@ -312,15 +335,36 @@ lazy val stagedRelease = (project in file("core/src/test"))
 
 import ReleaseTransformations._
 
-// Don't use sbt-release's cross facility	
+// Don't use sbt-release's cross facility
 releaseCrossBuild := false
 
-def crossReleaseStep(step: ReleaseStep): Seq[ReleaseStep] = {
+lazy val downVersionPySpark = taskKey[Unit]("Replace PySpark")
```
karenfeng, line 341: To future-proof this for PySpark 3.0.1+ and avoid hard-coding this, I think it'd be useful to change this to "changePysparkVersion" and accept a parameter so we can toggle versions more freely without adding a bespoke task per version.

build.sbt
karenfeng: I agree they won't be needed for `publishArtifacts` and `stagedRelease` - both only touch the Scala side.

build.sbt
```diff
@@ -312,15 +335,36 @@ lazy val stagedRelease = (project in file("core/src/test"))
...
+downVersionPySpark := {
+  "./downversion-pyspark.sh" !
```
karenfeng, line 343: You've checked in a `txt` I believe - can you double-check?

karenfeng (commented):
Thanks for working on this Kiavash! Looks good so far. Let me know when you're ready for another pass.

build.sbt
kianfar77: Done

build.sbt
kianfar77: I parametrized it to sparkOld. Assumption is that the environment.yml file has the sparkNew version of pyspark.

build.sbt
kianfar77: Great

build.sbt
kianfar77: I removed the sh file and inlined the script.

kianfar77:
> @kianfar77 I think we're missing the script to change the pyspark version?

I inlined the script.

build.sbt
```diff
@@ -7,10 +7,21 @@ import sbt.Keys._
 import sbt.librarymanagement.ModuleID
 import sbt.nio.Keys._
 
-lazy val scala212 = "2.12.8"
-lazy val scala211 = "2.11.12"
+lazy val scalaNew = "2.12.8"
```
henrydavidge, line 10: nit: I think it's better to keep descriptive names like 212 and 211. "new" and "old" tend to not age well.

build.sbt
```diff
@@ -7,10 +7,21 @@ import sbt.Keys._
...
+lazy val scalaOld = "2.11.12"
 
-lazy val sparkVersion = sys.env.getOrElse("SPARK_VERSION", "3.0.0")
+lazy val sparkNew = "3.0.0"
```
henrydavidge, line 13: likewise, I think we should use `spark3` and `spark2`

build.sbt
```diff
@@ -144,30 +161,30 @@ lazy val root = (project in file(".")).aggregate(core, python, docs)
...
+        "Only supported Scala versions are: " + Seq(scalaOld, scalaNew))
   }
 }
 
 lazy val core = (project in file("core"))
   .settings(
     commonSettings,
     functionGenerationSettings,
-    name := "glow",
+    name := s"glow_spark${majorVersion(sparkVersion.value)}_scala",
```
henrydavidge, line 176: are you sure we want the `_scala`? sbt usually adds that automatically. Also, I think it's a bit more common to use `-` instead of `_` in artifact names. See search results for example: https://mvnrepository.com/search?q=spark2.

henrydavidge (commented):
Had a few questions / minor comments

build.sbt
kianfar77: reverted

build.sbt
kianfar77: reverted

build.sbt
kianfar77: changed _ to -. 
I think it is better to keep scala. sbt adds _2.11 or _2.12 which will look confusing: glow spark3_2.11. Even the currently release artifacts are confusing glow_2.11. It looks as if 2.11 is version of glow or part of version of spark.

henrydavidge (approved):

PR #292, opened 2020-09-04
Fix BgenRowConverterSuite with AQE enabled

karenfeng:

We need to use `map(_.copy())` every time we use `queryExecution.toRdd`, as `InternalRow`s are re-used (see https://github.com/apache/spark/blob/76330e029540fa7ff4311c682135efd8bf0804e5/sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala#L121). Without this, our tests failed when adaptive query execution (AQE) was enabled. I fixed this bug in our tests and added a regression test.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

### DIFF
--- a/core/src/test/scala/io/projectglow/bgen/BgenRowConverterSuite.scala
+++ b/core/src/test/scala/io/projectglow/bgen/BgenRowConverterSuite.scala
@@ -17,6 +17,7 @@
 package io.projectglow.bgen
 
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
+import org.apache.spark.sql.SparkSession
 
 import io.projectglow.common.{BgenGenotype, BgenRow}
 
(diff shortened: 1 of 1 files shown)


core/src/test/scala/io/projectglow/bgen/BgenRowConverterSuite.scala
```diff
@@ -49,6 +50,7 @@ class BgenRowConverterSuite extends BgenConverterBaseTest {
     val vcfRows = vcfDf
       .queryExecution
       .toRdd
+      .map(_.copy())
```
henrydavidge, line 53: nit: can we just move the `converter.convert` here?

karenfeng: I don't believe we can - `converter` is not serializable.

henrydavidge: How about `.mapPartitions { it => val converter =... }`?

henrydavidge: The copy is just pretty confusing

PR #297, opened 2020-10-07
Re-initialize beta vector between genotypes in LogisticRegression

henrydavidge:

Previously we were not re-initializing the `beta` vector between genotypes for a given phenotype. This means that if one row results in a `NaN` beta, all successive rows will fail to converge.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/sql/expressions/FirthTest.scala
+++ b/core/src/main/scala/io/projectglow/sql/expressions/FirthTest.scala
@@ -161,6 +161,7 @@ class FirthNewtonArgs(numRows: Int, numCols: Int) {
 
   def initFromNullFit(nullFit: FirthNewtonArgs): Unit = {
     b(0 until nullFit.b.length) := nullFit.b
+    b(nullFit.b.length to -1)
   }
 }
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala/io/projectglow/sql/expressions/FirthTest.scala
```diff
@@ -161,6 +161,7 @@ class FirthNewtonArgs(numRows: Int, numCols: Int) {
 
   def initFromNullFit(nullFit: FirthNewtonArgs): Unit = {
     b(0 until nullFit.b.length) := nullFit.b
+    b(nullFit.b.length to -1) := 0d
```
henrydavidge, line 164: I couldn't reproduce the same failure mode for Firth logistic regression, but we should have the same re-initialization logic here.

kianfar77: I think for firth we `clear` before each fit because we are not caching and `clear` takes care of it.

kianfar77 (approved):

core/src/main/scala/io/projectglow/sql/expressions/FirthTest.scala
henrydavidge: Ah, yes

kianfar77:
Perhaps we should add a unit test for this as well?

PR #298, opened 2020-10-12
Disable gatk docs link checks

henrydavidge:

These pages are now gated by a captcha. It's unfortunate that our link checker doesn't work here since the GATK links are pretty important, but I don't think there's a way around.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


karenfeng (approved):
LGTM, thanks for taking care of this!

PR #300, opened 2020-10-21
Conversion from Hail MatrixTable to Glow DataFrame

karenfeng:

Adds interoperability between Hail and Glow by converting from Hail MatrixTable to Glow DataFrames.

Some of the input files were malformed, so I corrected them.

## How is this patch tested?
- [x] Unit tests
- [x] Integration tests
- [x] Manual tests

I performed manual timing for converting a MatrixTable to a Glow DF on a cluster with an i3.xlarge driver and 8 c4.2xlarge workers.

Samples | Variants | With sample IDs | Minutes to: from_matrix_table(mt) | Minutes to: df.write.format("noop") | Samples * variants / minute / core
-- | -- | -- | -- | -- | --
2504 | 1103547 | TRUE | 0.07 | 1.56 | 26521054
2504 | 1103547 | FALSE | 0.07 | 1.32 | 31106827
2504 | 81271745 | TRUE | 0.04 | 105 | 30273165
50000 | 100000 | TRUE | 0.13 | 2.27 | 32617771
500000 | 100000 | TRUE | 1.34 | 76.2 | 10075444
500000 | 504291 | TRUE | 1.35 | 246 | 15927929

For comparison, I ran similar tests using the Glow datasources:

Samples | Variants | With sample IDs | Data source | Minutes to: spark.read | Minutes to: df.write.format("noop") | Times faster
-- | -- | -- | -- | -- | -- | --
2504 | 1103547 | TRUE | VCF | 0.05 | 0.62 | 2.43408921
2504 | 1103547 | FALSE | VCF | 0.02 | 0.60 | 2.261814231
2504 | 1103547 | FALSE | BGEN | 0.01 | 0.77 | 1.782152793
2504 | 1103547 | TRUE | Delta | 0.04 | 0.56 | 2.714841579
2504 | 81271745 | TRUE | Delta | 0.02 | 21.10 | 4.972408515


python/glow/hail/functions.py
```diff
@@ -0,0 +1,183 @@
...
+from pyspark.sql import Column, DataFrame
+import pyspark.sql.functions as fx
+from typing import List, NoReturn, Optional
+from typeguard import check_argument_types, check_return_type
+
+
+def __get_sample_ids(col_key: StructExpression, include_sample_ids: bool) -> Optional[List[str]]:
+    assert check_argument_types()
+
+    has_sample_ids = isinstance(col_key.dtype,
```
henrydavidge, line 29: Why do we need the first check?

python/glow/hail/functions.py
```diff
@@ -0,0 +1,183 @@
...
+    # null filters means missing, [] filters means PASS
+    if 'filters' in row and row.filters.dtype == tset(tstr):
+        other_cols.append(
+            fx.expr("if(size(filters) = 0, array('PASS'), if(isnull(filters), array(), filters))").
+            alias("filters"))
+    # Rename info.* columns to INFO_*
+    if 'info' in row and isinstance(row.info.dtype, tstruct):
+        for f in row.info:
+            if row.info[f].dtype == tbool:
+                # FLAG INFO fields should be null if not present; Hail encodes them as false
```
henrydavidge, line 137: Do you think we should fix this in our VCF parser? Seems like Hail's behavior makes more sense.

python/glow/hail/functions.py
```diff
@@ -0,0 +1,183 @@
...
+                other_cols.append(
+                    fx.expr(f"if(`info.{f}` = false, null, `info.{f}`)").alias(f"INFO_{f}"))
+            else:
+                other_cols.append(fx.col(f"`info.{f}`").alias(f"INFO_{f}"))
+
+    assert check_return_type(other_cols)
+    return other_cols
+
+
+def __require_row_variant_w_struct_locus(mt: MatrixTable) -> NoReturn:
```
henrydavidge, line 147: Why should we require these fields?

python/glow/hail/tests/test_from_matrix_table.py
```diff
@@ -0,0 +1,171 @@
...
+        matching_fields = [f2 for f2 in s2_fields if f1.name == f2.name]
+        assert (len(matching_fields) == 1)
+        m = matching_fields[0]
+        if isinstance(m.dataType, ArrayType) and isinstance(m.dataType.elementType, StructType):
+            __compare_struct_types(f1.dataType.elementType, m.dataType.elementType, ignore_fields)
+        else:
+            assert f1.dataType == m.dataType
+
+
+def __compare_round_trip(spark,
```
henrydavidge, line 38: nit: This isn't really a round trip, right? Perhaps it would be better to say `assert_adapter_matches_reader` or `assert_adapter_matches_glow_reader`

henrydavidge (commented):
Looks really good! Most of my comments are fairly minor. I do think we should make the build more isolated, though.

.circleci/config.yml
```diff
@@ -156,6 +160,8 @@ jobs:
     <<: *setup_base
     steps:
       - checkout
+      - run: sudo apt-get update
+      - run: sudo apt-get -y install rsync
```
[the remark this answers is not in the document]
karenfeng: This is for the Hail setup; they use rsync in the Makefile.

python/glow/hail/functions.py
karenfeng: Whoops, good catch.

python/glow/hail/functions.py
karenfeng: I'm not sure if there is any MatrixTable without these fields that we could meaningfully convert to a Glow-compatible Spark DataFrame; this quickly signals that the input is misconfigured.

python/glow/hail/functions.py
karenfeng: We can do that; we'd also have to update the fast VCF reader.

python/glow/conftest.py
```diff
@@ -12,14 +12,16 @@
...
+import hail as hl
+from pyspark.sql import functions, Row
+import pytest
 
 
-@pytest.fixture(autouse=True, scope="module")
+@pytest.fixture(autouse=True, scope="function")
 def add_spark(doctest_namespace, spark):
     glow.register(spark)
+    hl.init(spark.sparkContext, idempotent=True, quiet=True)
```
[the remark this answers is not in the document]
karenfeng: This takes a pretty negligible amount of time.

build.sbt
```diff
@@ -202,6 +205,25 @@ def currentGitHash(dir: File): String = {
   ).!!.trim
 }
 
+lazy val setupHail = taskKey[Unit]("Set up Hail")
```
henrydavidge, line 208: nit: might be better to call this `installHail`

build.sbt
```diff
@@ -202,6 +205,25 @@ def currentGitHash(dir: File): String = {
...
+    "source $(conda info --base)/etc/profile.d/conda.sh &&" +
+    "conda create -y --name hail &&" +
+    "conda activate hail --stack &&" +
+    s"make -C hail/hail SCALA_VERSION=${scalaVersion.value} SPARK_VERSION=${sparkVersion.value} install-deps shadowJar wheel &&" +
+    s"pip install --no-deps hail/hail/build/deploy/dist/hail-${hailVersion.value}-py3-none-any.whl"
+  ) !
+}
+
+lazy val teardownHail = taskKey[Unit]("Tear down Hail")
+ThisBuild / teardownHail := {
```
henrydavidge, line 223: I think this should be `cleanHail` or `uninstallHail`. `teardown` usually refers to the decomissioning of an active process.

build.sbt
```diff
@@ -225,7 +247,12 @@ lazy val pythonSettings = Seq(
       baseEnv :+ "ARROW_PRE_0_15_IPC_FORMAT" -> "1"
     }
     val ret = Process(
-      Seq("pytest") ++ args,
+      Seq(
+        "/bin/bash",
+        "-c",
+        "source $(conda info --base)/etc/profile.d/conda.sh &&" +
```
henrydavidge, line 253: Is this pretty standard?

.circleci/config.yml
```diff
@@ -52,12 +53,12 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - conda-deps-v1-{{ checksum "python/environment.yml" }}
+            - conda-deps-v1-{{ checksum "python/conda_version.txt" }}-{{ checksum "python/environment.yml" }}
```
henrydavidge, line 56: Why do we have a dependency on the conda version now?

build.sbt
henrydavidge: Also, I think we talked about having a separate test key for tests that rely on Hail? We should make sure that we don't require Hail or Hail deps for core Glow code.

henrydavidge (commented):
The main code looks good to me now. Just some comments / clarifications on the build side. Also, once we finalize how to run Hail tests, we should document in the README.

build.sbt
karenfeng: This is the line added to `.bashrc`/`.bash_profile` when installing conda in non-batch mode (without `-b`) for initialization purposes; without this, we can't use commands such as `conda activate`.

.circleci/config.yml
karenfeng: I thought my issues were due to an outdated conda version, but turns out that was a red herring. I'll remove this.

build.sbt
karenfeng: I've broken this out into `hail/test`, which runs a version of `pytest` but within the `hail` conda environment. We'll still have to streamline how this fits into releases; if we're trying to separate it from core Glow code, it makes sense to keep it out of `root` and make a new cross-release step to install/test/uninstall Hail.

henrydavidge (approved):
Nice! LGTM

PR #301, opened 2020-10-23
Generate notebook sources

karenfeng:

Adds generated notebook sources for our documentation for easy diff-ability by converting from HTML to source code. Checks via CI that the generated sources are not out-of-date.

## How is this patch tested?
- [ ] Unit tests
- [x] Integration tests
- [x] Manual tests

To generate a notebook source yourself:
- Update your conda environment with the new dependencies
- Configure your Databricks CLI
- Run `gen-nb-src.py` with the input HTML

Example of CI catching failures: https://app.circleci.com/pipelines/github/projectglow/glow/1646/workflows/58e2deeb-7036-468e-8975-bcc3115a04a0/jobs/3878

### DIFF
--- a/docs/dev/gen-nb-src.py
+++ b/docs/dev/gen-nb-src.py
@@ -40,7 +40,7 @@ def main(html, cli_profile, workspace_tmp_dir):
 
     def run_cli_workspace_cmd(args):
         cmd = ['databricks', '--profile', cli_profile, 'workspace'] + args
-        subprocess.run(cmd, stderr=subprocess.STDOUT, capture_output=True)
+        subprocess.run(cmd, capture_output=True)
         subprocess.check_output(cmd, stderr=subprocess.STDOUT)
 
     work_dir = os.path.join(workspace_tmp_dir, str(uuid.uuid4()))
--- a/docs/dev/gen-nb-src.py
+++ b/docs/dev/gen-nb-src.py
@@ -41,7 +41,7 @@ def main(html, cli_profile, workspace_tmp_dir):
     def run_cli_workspace_cmd(args):
         cmd = ['databricks', '--profile', cli_profile, 'workspace'] + args
         subprocess.run(cmd, capture_output=True)
-        subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+        # subprocess.check_output(cmd, stderr=subprocess.STDOUT)
 
     work_dir = os.path.join(workspace_tmp_dir, str(uuid.uuid4()))
     workspace_path = os.path.join(work_dir, rel_path)
--- a/docs/dev/gen-nb-src.py
+++ b/docs/dev/gen-nb-src.py
@@ -40,7 +40,8 @@ def main(html, cli_profile, workspace_tmp_dir):
 
     def run_cli_workspace_cmd(args):
         cmd = ['databricks', '--profile', cli_profile, 'workspace'] + args
-        subprocess.run(cmd, capture_output=True)
+        res = subprocess.run(cmd, capture_output=True)
+        raise ValueError(res)
         # subprocess.check_output(cmd, stderr=subprocess.STDOUT)
 
     work_dir = os.path.join(workspace_tmp_dir, str(uuid.uuid4()))
--- a/docs/dev/gen-nb-src.py
+++ b/docs/dev/gen-nb-src.py
@@ -21,22 +21,22 @@ SOURCE_EXTS = ['scala', 'py', 'r', 'sql']
 @click.command()
 @click.option('--html', required=True, help='Path of the HTML notebook.')
 @click.option('--cli-profile', default='docs-ci', help='Databricks CLI profile name.')
-@click.option('--workspace-tmp-dir', default='/tmp/glow-docs-ci', help='Temp workspace dir for import/export.')
+@click.option('--workspace-tmp-dir', default='/tmp/glow-docs-ci', help='Base workspace dir; a temporary directory will be generated under this for import/export.')
 def main(html, cli_profile, workspace_tmp_dir):
     assert os.path.commonpath([NOTEBOOK_DIR, html]) == NOTEBOOK_DIR, \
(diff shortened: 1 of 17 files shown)


docs/dev/gen-nb-src.py
```diff
@@ -0,0 +1,59 @@
...
+        for ext in SOURCE_EXTS:
+            source_path = os.path.join(SOURCE_DIR, rel_path + "." + ext)
+            if os.path.exists(source_path):
+                os.remove(source_path)
+                print("\tDeleted {}.".format(source_path))
+        return
+
+    print("Generating source file for {} under {} ...".format(html, SOURCE_DIR))
+
+    def run_cli_workspace_cmd(args):
```
henrydavidge, line 41: Why is this a nested definition?

docs/dev/gen-nb-src.py
```diff
@@ -0,0 +1,59 @@
...
+SOURCE_EXTS = ['scala', 'py', 'r', 'sql']
+
+
+@click.command()
+@click.option('--html', required=True, help='Path of the HTML notebook.')
+@click.option('--cli-profile', default='docs-ci', help='Databricks CLI profile name.')
+@click.option('--workspace-tmp-dir', default='/tmp/glow-docs-ci', help='Temp workspace dir for import/export.')
+def main(html, cli_profile, workspace_tmp_dir):
+    assert os.path.commonpath([NOTEBOOK_DIR, html]) == NOTEBOOK_DIR, \
+        "HTML notebook must be under {} but got {}.".format(NOTEBOOK_DIR, html)
```
henrydavidge, line 27: nit: prefer f strings for interpolation

docs/dev/gen-nb-src.py
```diff
@@ -0,0 +1,59 @@
...
+    rel_path = os.path.splitext(os.path.relpath(html, NOTEBOOK_DIR))[0]
+
+    if not os.path.exists(html):  # html notebook was deleted
+        print("{} does not exist. Deleting the companion source file...".format(html))
+        for ext in SOURCE_EXTS:
+            source_path = os.path.join(SOURCE_DIR, rel_path + "." + ext)
+            if os.path.exists(source_path):
+                os.remove(source_path)
+                print("\tDeleted {}.".format(source_path))
```
henrydavidge, line 36: Why the tab?

docs/dev/gen-nb-src.py
```diff
@@ -0,0 +1,59 @@
...
+import uuid
+
+NOTEBOOK_DIR = 'docs/source/_static/notebooks'
+SOURCE_DIR = 'docs/source/_static/zzz_GENERATED_NOTEBOOK_SOURCE'
+SOURCE_EXTS = ['scala', 'py', 'r', 'sql']
+
+
+@click.command()
+@click.option('--html', required=True, help='Path of the HTML notebook.')
+@click.option('--cli-profile', default='docs-ci', help='Databricks CLI profile name.')
```
henrydavidge, line 23: Might it make more sense to rely on the databricks CLI's default profile?

docs/dev/gen-nb-src.py
```diff
@@ -0,0 +1,59 @@
...
+@click.option('--workspace-tmp-dir', default='/tmp/glow-docs-ci', help='Temp workspace dir for import/export.')
```
henrydavidge, line 24: We should probably generate a temporary directory if one is not provided. In fact, maybe we should always just make a new one?

henrydavidge (commented):
Had a few comments on the script. This is cool!

docs/dev/gen-nb-src.py
karenfeng: I found this useful during manual testing, as I already have a default Databricks CLI profile and wanted to test this on the same shard as the CI.

docs/dev/gen-nb-src.py
karenfeng: Actually, we do generate a temporary work directory under this; this serves as the root directory. I'll clarify in the description.

docs/dev/gen-nb-src.py
karenfeng: We printed a line about the HTML file before this; here, we add a line for all associated generated source files (although reasonably, there should only be 1).

docs/dev/gen-nb-src.py
karenfeng: This is just so we can leverage `cli_profile` from the main args, but I can split this out if it's clearer that way.

docs/dev/gen-nb-src.py
henrydavidge: I mean that it makes sense to have this argument, but the default value should be the default profile. If you're running locally, I don't think there's any reason to expect a profile called `docs-ci` to exist. We can pass the `docs-ci` argument from our CI script.

docs/dev/gen-nb-src.py
henrydavidge: I think that would be a bit cleaner.

python/environment.yml
```diff
@@ -18,6 +18,8 @@ dependencies:
   - sphinx
   - sphinx_rtd_theme
   - pip:
+    - click==7.1.1 # Docs notebook source generation
```
henrydavidge, line 21: Can we install either of these from conda?

henrydavidge (commented):
Few small comments

python/environment.yml
```diff
@@ -1,20 +1,23 @@
...
-  - numpy=1.17.4
-  - scipy=1.4.1
-  - scikit-learn=0.22.1
   - nomkl # Skip MKL for local development
+  - numpy=1.17.4
   - pandas=0.25.3
   - pip
   - pyarrow=0.15.1
   - pyspark=3.0.0
+  - python=3.7
```
henrydavidge, line 16: I think it's actually standard to list python first since it is pretty important :).

henrydavidge (approved):
Nice!

PR #303, opened 2020-10-30
Create BGEN reader option to add hard calls

karenfeng:

Adds an option to the BGEN reader to set the genotype calls based on the posterior probabilities with a hard call threshold (the posterior probabilities are still present). Calls are set to missing for non-diploids; this is based on the logic for the existing `hard_calls` function. To match the behavior from qctool, calls for unphased genotypes are now ascending.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

I performed manual testing to materialize a DataFrame with/without hard calls on a cluster with an i3.xlarge driver and 2 i3.xlarge workers, and observed a runtime increase from 3.16 minutes to 4.36 minutes.


core/src/main/scala/io/projectglow/bgen/BgenRowToInternalRowConverter.scala
```diff
@@ -23,14 +23,16 @@ import org.apache.spark.sql.types.{ArrayType, StructType}
...
+import io.projectglow.sql.expressions.HardCalls
 import io.projectglow.sql.util.RowConverter
 
 /**
  * Converts [[BgenRow]]s into [[InternalRow]] with a given required schema. During construction,
  * this class will throw an [[IllegalArgumentException]] if any of the fields in the required
  * schema cannot be derived from a BGEN record.
  */
-class BgenRowToInternalRowConverter(schema: StructType) extends GlowLogging {
+class BgenRowToInternalRowConverter(schema: StructType, hardCallsThreshold: Option[Double])
```
henrydavidge, line 34: Why is this an option? Seems like we could always pass in the default or the user-provided override.

core/src/main/scala/io/projectglow/bgen/BgenRowToInternalRowConverter.scala
```diff
@@ -71,21 +74,40 @@ class BgenRowToInternalRowConverter(schema: StructType) extends GlowLogging {
...
 
-  private def makeGenotypeConverter(gSchema: StructType): RowConverter[BgenGenotype] = {
+  private def makeGenotypeConverter(
+      gSchema: StructType,
+      hardCallsThreshold: Option[Double]): RowConverter[(Int, BgenGenotype)] = {
     val functions = gSchema.map { field =>
-      val fn: RowConverter.Updater[BgenGenotype] = field match {
+      val fn: RowConverter.Updater[(Int, BgenGenotype)] = field match {
         case f if structFieldsEqualExceptNullability(f, sampleIdField) =>
           (g, r, i) => {
```
henrydavidge, line 83: Can you expand the signature so that it looks like `((numAlleles, g), r, i)`? The `._1` and `._2` are hard to read.

core/src/main/scala/io/projectglow/bgen/BgenSchemaInferrer.scala
```diff
@@ -28,24 +28,33 @@ import io.projectglow.sql.util.SerializableConfiguration
...
+ *   `sampleId` is added to the genotype field.
+ * - If a hard call threshold is specified, `calls` is added to the genotype field.
  */
 object BgenSchemaInferrer {
   def inferSchema(
       spark: SparkSession,
       files: Seq[FileStatus],
       options: Map[String, String]): StructType = {
+    val hasSampleIds = includeSampleIds(spark, files, options)
+    val hasHardCalls = options.contains(BgenOptions.HARD_CALL_THRESHOLD)
```
henrydavidge, line 41: Hm, I feel like we should include hard calls by default. AFAIK, every user so far has wanted hard calls when reading BGEN files. We could add a separate flag to disable, or accept -1 here.

henrydavidge (commented):
Just a few comments. Do you know if there's a convention regarding order of calls for heterogeneous unphased samples in other tools?

karenfeng:
As far as I can tell, the default is always to emit unphased genotype calls in ascending order (eg. in VCFs).

henrydavidge (approved):
LGTM!

PR #309, opened 2020-11-19
Support an offset in linear regression

henrydavidge:

Users can now provide an offset to the pandas based linear regression function. This offset takes the same form as the return from GloWGR step 1. The offset can either be per sample or per contig and sample.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


python/glow/gwas/linear_regression.py
```diff
@@ -72,31 +80,79 @@ def linear_regression(genotype_df: DataFrame,
...
-                                           phenotype_df.columns.to_series().astype('str'),
-                                           values_column)
+            yield _linear_regression_dispatch(pdf, Y_state, Y_mask, Q, dof,
+                                              phenotype_df.columns.to_series().astype('str'),
+                                              values_column)
 
     return genotype_df.mapInPandas(map_func, result_struct)
 
 
+def _indexes_have_same_elements(idx1: pd.Index, idx2: pd.Index) -> bool:
```
kianfar77, line 121: perhaps not call this function `_indexes...` as you use it for non-index comparisons.

kianfar77 (approved):
Looks great. Just a nit.

PR #310, opened 2020-11-19
Add documentation for Hail interoperation

karenfeng:

Add docs for the Hail interop function `from_matrix_table`, which will be available in 0.7.

Also fixes a bug in the aforementioned function; previously, missing names were represented as an empty array; this should be null.

## How is this patch tested?
- [ ] Unit tests
- [x] Integration tests
- [ ] Manual tests

Includes docs-side testing in `hail/test`. This is chained to our Python-side testing to leverage the Spark-Hail setup; otherwise, we'd have to copy `python/glow/hail/conftest.py` to the docs.


build.sbt
```diff
@@ -301,11 +305,11 @@ lazy val python =
     )
     .dependsOn(core % "test->test")
 
-lazy val hail = (project in file("python/glow/hail"))
+lazy val hail = (project in file("hail"))
   .settings(
     pythonSettings,
     test in Test := {
-      hailtest.toTask(" --doctest-modules python/glow/hail/").value
+      hailtest.toTask(" --doctest-modules python/glow/hail/ docs/source/etl/hail.rst").value
```
henrydavidge, line 312: nit: make this into a variable `hailPaths` in case we have more hail related docs in the future

python/glow/hail/functions.py
```diff
@@ -156,6 +154,8 @@ def from_matrix_table(mt: MatrixTable, include_sample_ids: bool = True) -> DataF
     """
     Converts a Hail MatrixTable to a Glow DataFrame.
 
+    Requires that the MatrixTable rows contain locus and alleles fields.
+
     Args:
         mt : The Hail MatrixTable to convert
         include_sample_ids : If true, include sample IDs in the Glow DataFrame
```
henrydavidge, line 161: nit: could you document why you might want this to be false?

python/glow/hail/functions.py
```diff
@@ -156,6 +154,8 @@ def from_matrix_table(mt: MatrixTable, include_sample_ids: bool = True) -> DataF
     """
     Converts a Hail MatrixTable to a Glow DataFrame.
```
henrydavidge, line 155: Could you describe what the schema will look like? How is it translated from the Hail schema?

henrydavidge (approved):

PR #311, opened 2020-11-24
Ignore covariates in binary WGR transform for linear response

karenfeng:

Changes how we handle covariates passed into binary GloWGR's transform function. Rather than handling it as a special case in `transform_loco`, pushes the replacement of the covariate matrix down to the transform function.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

### DIFF
--- a/python/glow/wgr/linear_model/logistic_model.py
+++ b/python/glow/wgr/linear_model/logistic_model.py
@@ -19,6 +19,7 @@ from pyspark.sql.functions import pandas_udf, PandasUDFType
 import pyspark.sql.functions as f
 from typeguard import typechecked
 from typing import Any, Dict, List
+import warnings
 from glow.logging import record_hls_event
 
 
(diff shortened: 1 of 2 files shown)


python/glow/wgr/linear_model/logistic_model.py
```diff
@@ -163,6 +163,9 @@ def reduce_block_matrix(self, blockdf: DataFrame, labeldf: pd.DataFrame,
         transform_key_pattern = ['sample_block', 'label']
 
         if response == 'linear':
+            if not covdf.empty:
+                print('Ignoring covariates for linear response')
```
henrydavidge, line 167: It might make more sense to emit a warning here. Also, we should explain why in the function documentation for the user facing method.

henrydavidge (approved):

PR #312, opened 2020-11-24
Accept different datatypes, values column expression for linear regression

henrydavidge:
Two usability improvements. I checked to see if using float32s makes any performance difference. It doesn't seem to make much of a difference with openblas, although it did decrease peak memory usage some.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/python/glow/gwas/linear_regression.py
+++ b/python/glow/gwas/linear_regression.py
@@ -54,7 +54,7 @@ def linear_regression(genotype_df: DataFrame,
         - `effect`: The effect size estimate for the genotype
         - `stderror`: The estimated standard error of the effect
         - `tvalue`: The T statistic
-        - `pvalue`: P value estimed from a two sided T-test
+        - `pvalue`: P value estimated from a two sided T-test
         - `phenotype`: The phenotype name as determined by the column names of `phenotype_df`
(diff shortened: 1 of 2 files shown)


python/glow/gwas/linear_regression.py
```diff
@@ -63,26 +70,47 @@ def linear_regression(genotype_df: DataFrame,
...
+        StructField('stderror', sql_type),
+        StructField('tvalue', sql_type),
+        StructField('pvalue', sql_type),
         StructField('phenotype', StringType())
     ]
-    fields = [field
-              for field in genotype_df.schema.fields if field.name != values_column] + result_fields
+    fields = [
+        field for field in genotype_df.schema.fields
+        if field.name not in set([_VALUES_COLUMN_NAME, _GENOTYPES_COLUMN_NAME])
```
kianfar77, line 102: Is `_GENOTYPES_COLUMN_NAME` needed here? The column is dropped above.

kianfar77 (approved):
LGTM subject to a few minor comments. Can you also mention dropping of the genotypes column in the PR description?

PR #314, opened 2020-12-02
Add docs for pandas based linear regression; misc doc improvements

henrydavidge:
The docs for linear regression should be self explanatory!

I also cleaned up how we reference different subpackages. Previously, we imported `functions`, `glow`, and some things from `wgr` into the top level `glow` module. I maintained those imports, but also redefined `__all__`` to enumerate the functions from these submodules.

Similarly, I changed the `__init__.py` files for the `gwas` and `wgr` submodules so that they appear to contain all of the classes / functions defined by all their submodules.

To access the different submodules, you can use the following import patterns:
```python
import glow
glow.register
glow.transform
glow.genotype_states
...
glow.gwas.linear_regression
glow.wgr.RidgeModel

# Hail module has to be imported separately since it's not a required dep
import glow.hail
glow.hail.from_matrix_table
```

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


python/glow/gwas/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .linear_regression import *
+__all__ = ['linear_regression']
```
henrydavidge, line 2: This is a way to accomplish what I mentioned about making the docs show up as `glow.gwas.linear_regression` instead of `glow.gwas.linear_regression.linear_regression`.

python/glow/wgr/linear_model/ridge_model.py
```diff
@@ -307,11 +307,11 @@ def transform_loco(self,
             sample_blocks : Dict containing a mapping of sample_block ID to a list of corresponding sample IDs
             modeldf : Spark DataFrame produced by the RidgeRegression fit method, representing the reducer model
             cvdf : Spark DataFrame produced by the RidgeRegression fit method, containing the results of the cross
-            validation routine.
+                validation routine.
```
henrydavidge, line 310: These indentation changes fix parameter formatting in the docs.

docs/source/environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - bioconda
   - conda-forge
 dependencies:
-  - hail>=0.2.58
+  - python=3.7
```
henrydavidge, line 6: `linear_regression` is not compatible with python3.6, which the environment installed by default

docs/source/environment.yml
```diff
@@ -13,5 +13,7 @@ dependencies:
   - sphinx_rtd_theme
   - typeguard=2.9.1
   - pip:
+      - hail>=0.2.58
```
henrydavidge, line 16: The bioconda distribution does not seem well maintained. They don't have a version for python 3.7.

docs/source/environment.yml
```diff
@@ -13,5 +13,7 @@ dependencies:
...
+      - sphinx-autobuild==2020.9.1
```
henrydavidge, line 17: Necessary to auto rebuild docs

python/glow/wgr/functions.py
```diff
@@ -19,6 +19,8 @@
 from typeguard import check_argument_types, check_return_type
 from typing import Dict, List
 
+__all__ = ['get_sample_ids', 'block_variants_and_samples', 'reshape_for_gwas']
```
karenfeng, line 22: We probably need a newline before each requirements list; I'm not seeing a formatted list.

python/glow/__init__.py
```diff
@@ -12,7 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from glow.glow import *
-from glow.functions import *
-from glow.hail import *
-from glow.wgr import *
+__all__ = []
```
karenfeng, line 15: Is this something that we have to manually fill in for any public functions?

karenfeng (commented):
When running `make livehtml` with the new conda env, I'm seeing the following errors:

```
WARNING: error while formatting arguments for glow.wgr.LogisticRegression: type object 'NDArray' has no attribute '_special'
```

```
/Users/karen.feng/glow/docs/source/blogs/release-0-4-0-blog/release-0-4-0-blog.rst:226: WARNING: undefined label: pyspark_functions (if the link has no caption the label must precede a section header)
```

henrydavidge:
@karenfeng I actually don't see the `_special` error. Maybe you need to recreate the environment?

Fixed the `pyspark_functions` thing.

docs/source/tertiary/regression-tests.rst
```diff
@@ -1,8 +1,8 @@
 .. _gwas:
 
-==============================================
-Genome-wide Association Study Regression Tests
-==============================================
+======================================================
+GloWGR: Genome-wide Association Study Regression Tests
```
kianfar77, line 4:
```suggestion
GloWGR: Genome-Wide Association Study (GWAS) Regression Tests
```

docs/source/tertiary/regression-tests.rst
```diff
@@ -27,18 +29,17 @@ genome-wide association studies (GWAS).
 Linear regression
 =================
 
-``linear_regression_gwas`` performs a linear regression association test optimized for performance
+``linear_regression`` performs a linear regression association test optimized for performance
```
kianfar77, line 32: Can we add some statements here as to what goes into the function and what comes out at a very high level? Also emphasis on the fact that multiple phenotypes can be processed in parallel is important. I know some concepts are explained in the API but I feel this paragraph needs some more high-level explanation to connect with the reader.

docs/source/tertiary/regression-tests.rst
```diff
@@ -47,97 +48,61 @@ Example
...
 
-Implementation details
-----------------------
+.. note::
 
-The linear regression model is fit using the QR decomposition. For performance, the QR decomposition
-of the covariate matrix is computed once and reused for each (``genotypes``, ``phenotypes``) pair.
+  Glow also includes a SQL-based function for performing linear regression. However, this function
+  only processes one phenotype at time, and so performs more slowly than the batch linear regression function
+  documented above. To read more about the SQL-based function, see the docs for :func:`glow.functions.linear_regression_gwas`.
```
kianfar77, line 105:
```suggestion
  documented above. To read more about the SQL-based function, see  :func:`glow.functions.linear_regression_gwas`.
```

python/glow/gwas/lin_reg.py
```diff
@@ -32,30 +32,54 @@ def linear_regression(genotype_df: DataFrame,
     In each Spark task, we project the covariates out of a block of genotypes and then compute the regression statistics for each phenotype,
     taking into account the distinct missingness patterns of each phenotype.
```
kianfar77, line 34: Can we add a less technical version of this paragraph to the Linear Regression doc page to echo the power of implementation?

python/glow/gwas/lin_reg.py
```diff
@@ -32,30 +32,54 @@ def linear_regression(genotype_df: DataFrame,
     In each Spark task, we project the covariates out of a block of genotypes and then compute the regression statistics for each phenotype,
```
kianfar77, line 32: Can we give a title of "Details of implementation" to this paragraph?

kianfar77, line 32: I would change "The implementation is based on regenie: https://www.biorxiv.org/content/10.1101/2020.06.19.162354v2" to "the implementation is a distributed version of the method used at regenie..." or something like that.

kianfar77 (changes requested):
Thanks. Some comments. The simplification of import process is beautiful.

python/glow/wgr/functions.py
karenfeng: Oh sorry, in the docstrings below. I just couldn't comment on the appropriate lines, eg:

```
 Requires that:
        - Each variant row has the same number of values
        - The number of values per row matches the number of sample IDs
```

These aren't actually made into a formatted list because I think there needs to be a leading newline.

docs/source/tertiary/regression-tests.rst
```diff
@@ -31,8 +31,14 @@ Linear regression
 =================
 
 ``linear_regression`` performs a linear regression association test optimized for performance
-in a GWAS setting. It uses vectorized linear algebra libraries to process multiple phenotypes
-efficiently.
+in a GWAS setting. You provide a Spark DataFrame containing the genetic data and Pandas DataFrame's
```
kianfar77, line 34: nit: I dont think we need `'` in `DataFrame's`

kianfar77 (approved):
LGTM. just a nit. thanks.

PR #316, opened 2020-12-10
Pandas based logistic regression

henrydavidge:
- Moves logic shared by pandas based linear and logistic regression to a common file
- Adds scaffolding for a pandas based logistic regression test with fallback logic for potentially significant variants
- Implements a fast multi-pheno, multi-geno score test

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


python/glow/gwas/logistic_regression.py
```diff
@@ -0,0 +1,185 @@
...
+fallback_none = 'none'
+
+
+@typechecked
+def logistic_regression(
+        genotype_df: DataFrame,
+        phenotype_df: pd.DataFrame,
+        covariate_df: pd.DataFrame = pd.DataFrame({}),
+        offset_df: pd.DataFrame = pd.DataFrame({}),
+        # TODO: fallback is probably not the best name
```
karenfeng, line 26: In addition to `fallback` (I propose `correction` as an alternative name), we should expose a parallel to `pvalue_threshold` below which we perform the correction.

python/glow/gwas/logistic_regression.py
```diff
@@ -0,0 +1,185 @@
...
+                        is provided, the expression should return a numeric array type.
+        dt : The numpy datatype to use in the linear regression test. Must be `np.float32` or `np.float64`.
+    '''
+
+    gwas_fx._check_spark_version(genotype_df.sql_ctx.sparkSession)
+    gwas_fx._validate_covariates_and_phenotypes(covariate_df, phenotype_df, is_binary=True)
+    sql_type = gwas_fx._regression_sql_type(dt)
+    genotype_df = gwas_fx._prepare_genotype_df(genotype_df, values_column, sql_type)
+    result_fields = [
+        # TODO: Probably want to put effect size and stderr here for approx-firth
```
karenfeng, line 66: I think we can still calculate effect size and stderr without the corrections, right? As in: https://github.com/rgcgithub/regenie/blob/247483cd5617f048682062553265837c2b95d6ee/src/Data.cpp#L2456

karenfeng (commented):
I left a couple of high-level questions I had when writing a first cut for the approximate firth correction.

python/glow/gwas/logistic_regression.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def _logistic_residualize(X: NDArray[(Any, Any), Float], C: NDArray[(Any, Any), Float],
+                          Y_mask: NDArray[(Any, Any), Float], gamma: NDArray[(Any, Any), Float],
+                          inv_CtGammaC: NDArray[(Any, Any), Float]) -> NDArray[(Any, Any), Float]:
+    '''
+    Residualize the genotype vectors given the null model predictions.
+    X_res = X - C(C.T gamma C)^-1 C.T gamma X
+    '''
+    X_hat = gwas_fx._einsum('ic,pcd,ds,ps,sg,sp->igp', C, inv_CtGammaC, C.T, gamma, X, Y_mask)
```
kianfar77, line 150: is `inv_CtGammaC` three-dimensional? Not sure why `pcd` is used as subscript.

python/glow/gwas/logistic_regression.py
```diff
@@ -0,0 +1,185 @@
...
+
+@dataclass
+class LogRegState:
+    inv_CtGammaC: NDArray[(Any, Any), Float]
+    gamma: NDArray[(Any, Any), Float]
+    Y_res: NDArray[(Any, Any), Float]
+
+
+@typechecked
+def _assemble_log_reg_state(
```
kianfar77, line 125: nit: you use `create` for linear regression and `assemble` here. Perhaps better to use same verb in both.

python/glow/gwas/logistic_regression.py
```diff
@@ -0,0 +1,185 @@
...
+        Y_mask: NDArray[(Any, Any), Float]) -> LogRegState:
+    Y_pred = np.row_stack([
+        _logistic_null_model_predictions(
+            Y[:, i], C, Y_mask[:, i],
+            offset_df.iloc[:, i].to_numpy() if offset_df is not None else None)
+        for i in range(Y.shape[1])
+    ])
+    gamma = Y_pred * (1 - Y_pred)
+    CtGammaC = C.T @ (gamma[:, :, None] * C)
+    CtGammaC_inv = np.linalg.inv(CtGammaC)
```
kianfar77, line 139: nit: some places `CtGammaC_inv` is used and some other places `inv_CtGammaC`, can we fix on one?

python/glow/gwas/logistic_regression.py
```diff
@@ -0,0 +1,185 @@
...
+                    If two levels, the level 0 index should be the same as the `phenotype_df`, and the level 1 index
+                    should be the contig name. The two level index scheme allows for per-contig offsets like
+                    LOCO predictions from GloWGR.
+        fallback : Which test to use for variants that meet a significance threshold for the score test
+        fit_intercept : Whether or not to add an intercept column to the covariate DataFrame
+        values_column : A column name or column expression to test with linear regression. If a column name is provided,
+                        `genotype_df` should have a column with this name and a numeric array type. If a column expression
+                        is provided, the expression should return a numeric array type.
+        dt : The numpy datatype to use in the linear regression test. Must be `np.float32` or `np.float64`.
+    '''
```
kianfar77, line 59: Can we add `Returns` descriptions?

kianfar77 (changes requested):
Looks nice! I had some comments.

python/glow/gwas/logistic_regression.py
henrydavidge: I saw that they compute `bhat`, but I don't understand what they actually represent. Based on the regenie paper and other resources, it seems that "effect size" universally corresponds to the maximum likelihood coefficient for the genotype feature. If that's the case, how could you know the effect size without fitting a model?

python/glow/gwas/linear_regression.py
```diff
@@ -113,50 +86,23 @@ def linear_regression(genotype_df: DataFrame,
     np.nan_to_num(Y, copy=False)
     _residualize_in_place(Y, Q)
 
-    if not offset_df.empty:
```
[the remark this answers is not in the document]
henrydavidge: We check that they have the same row index, which is actually more strict.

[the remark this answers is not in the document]
kianfar77: My bad. I just saw the columns comparison.

python/glow/gwas/logistic_regression.py
```diff
@@ -0,0 +1,185 @@
...
+    # nptying can't handle optional NDArrays, so don't verify the type for offset
+    offset
+) -> NDArray[Any, Float]:
+    if offset is not None:
+        offset = offset[y_mask]
+    model = sm.GLM(y[y_mask],
+                   X[y_mask],
+                   family=sm.families.Binomial(),
+                   offset=offset,
+                   missing='ignore')
```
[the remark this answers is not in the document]
henrydavidge: Good catch! It looks like if you specify an invalid option, statsmodels defaults to `none`, which is actually what I want. I will change to `none`.

python/glow/gwas/logistic_regression.py
karenfeng: I believe that effect size here simply refers to the difference between means, which is similar to the t-test stat (without the scaling).

python/glow/gwas/log_reg.py
```diff
@@ -23,8 +23,8 @@ def logistic_regression(
         phenotype_df: pd.DataFrame,
         covariate_df: pd.DataFrame = pd.DataFrame({}),
         offset_df: pd.DataFrame = pd.DataFrame({}),
-        # TODO: fallback is probably not the best name
-        fallback: str = 'none',  # TODO: Make approx-firth default
+        correction: str = 'none',  # TODO: Make approx-firth default
```
kianfar77, line 26: nit: I think the term `correction` is misleading for this argument. Something like `alternative` may make more sense.

kianfar77 (approved):
Looks great! Just a couple of nits.

python/glow/gwas/log_reg.py
karenfeng: Correction is the term used in the regenie paper to refer to Firth/SPA.

karenfeng (commented):
As discussed offline, can you set a numpy random seed? Otherwise, we can end up with test sets with perfect separation (which we should probably have as well, but only when we expect it).

PR #317, opened 2020-12-18
Address edge case with tabix offsets and partitioned files

karenfeng:

Fixes the logic for parsing whether a partition file overlaps with the tabix offsets. In particular, does not include a partitioned file as overlapping if it does not include the start position of the last tabix offset (even if it's within 0xFFFF), as it only contains the tail of the block and will trigger a failure when trying to read the GZIP block header.

During this exploration, we also learned that the BGZF codec seeks to the first BGZF header and reads through to the end of any block it has already started, whether or not it is within the limits of the partitioned file: https://github.com/apache/hadoop/blob/d4c8d4d4d203c934e8074b31289a28724c0842cf/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/input/CompressedSplitLineReader.java#L46.

However, if there is no header in the partitioned file, the codec explodes. This PR adds an extra check that the partitioned file contains any of the desired block start positions; if not, they are not returned.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

Manual testing has verified that this resolves the edge case.

### DIFF
--- a/core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
+++ b/core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
@@ -489,10 +489,10 @@ object TabixIndexHelper extends GlowLogging {
               }
             val blockRangeStart = Math.max(file.start, minOverBlocks >> 16) // Shift 16 bits to get
             // file offset of the bin from bgzipped virtual file offset.
-            val blockRangeEnd = Math.min(file.start + file.length, (maxOverBlocks >> 16) + 0xFFFF)
+            val blockRangeEnd = Math.min(file.start + file.length, maxOverBlocks >> 16)
             // 0xFFFF is the maximum possible length of an uncompressed bin.
             if (blockRangeStart <= blockRangeEnd) {
-              Some((blockRangeStart, blockRangeEnd))
+              Some((blockRangeStart, Math.min(file.start + file.length, blockRangeEnd + 0xFFFF)))
             } else {
               None
             }
--- a/core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
+++ b/core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
@@ -482,16 +482,23 @@ object TabixIndexHelper extends GlowLogging {
           if (offsetList.isEmpty) {
             None
           } else {
-            val (minOverBlocks, maxOverBlocks) = offsetList
-              .foldLeft((offsetList(0).getStartPosition, offsetList(0).getEndPosition)) {
-                case ((myStart, myEnd), e) =>
-                  (Math.min(myStart, e.getStartPosition), Math.max(myEnd, e.getEndPosition))
-              }
-            val blockRangeStart = Math.max(file.start, minOverBlocks >> 16) // Shift 16 bits to get
-            // file offset of the bin from bgzipped virtual file offset.
-            val blockRangeEnd = Math.min(file.start + file.length, maxOverBlocks >> 16)
-            // 0xFFFF is the maximum possible length of an uncompressed bin.
-            if (blockRangeStart <= blockRangeEnd) {
+            // Do not return any files without a BGZIP header
+            var fileContainsBlockStart = false
+            // Shift 16 bits to get file offset of the bin from bgzipped virtual file offset.
+            val firstStart, firstLast =
+              (offsetList(0).getStartPosition >> 16, offsetList(0).getEndPosition >> 16)
+            val (minOverBlocks, maxOverBlocks) = offsetList.foldLeft((firstStart, firstLast)) {
+              case ((aStart, aEnd), o) =>
+                val bStart = o.getStartPosition >> 16
+                val bEnd = o.getEndPosition >> 16
+                if (!fileContainsBlockStart) {
+                  fileContainsBlockStart = (file.start <= bStart) && (file.start + file.length >= bStart)
+                }
+                (Math.min(aStart, bStart), Math.max(aEnd, bStart))
+            }
+            val blockRangeStart = Math.max(file.start, minOverBlocks)
+            val blockRangeEnd = Math.min(file.start + file.length, maxOverBlocks)
+            if (blockRangeStart <= blockRangeEnd && fileContainsBlockStart) {
               Some((blockRangeStart, Math.min(file.start + file.length, blockRangeEnd + 0xFFFF)))
             } else {
               None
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
```diff
@@ -489,10 +489,10 @@ object TabixIndexHelper extends GlowLogging {
...
             // file offset of the bin from bgzipped virtual file offset.
-            val blockRangeEnd = Math.min(file.start + file.length, (maxOverBlocks >> 16) + 0xFFFF)
+            val blockRangeEnd = Math.min(file.start + file.length, maxOverBlocks >> 16)
             // 0xFFFF is the maximum possible length of an uncompressed bin.
             if (blockRangeStart <= blockRangeEnd) {
-              Some((blockRangeStart, blockRangeEnd))
+              Some((blockRangeStart, Math.min(file.start + file.length, blockRangeEnd + 0xFFFF)))
             } else {
               None
             }
```
kianfar77, line 498: Can you let me know what changes with this fix? It seems to me the final result at the end of line 498 will be the same before and after your changes. I could not immediately see this in your unit tests either.

kianfar77 (commented):
Thanks @karenfeng for working on this. I posted a question in the review. But in addition to that, I do not see the reason why the logic in ``getFileRangeToRead`` should  change. It seems to me that splitter for BGZF must split exactly at block breaks so that start of each partition is at the start of a block and if so this problem we have should not arise. The fact that this is not the case in this file means there is either a problem in bgzipping that file or indexing it or there is a bug in the splitter in this bgzfenhancedcodec.

core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
karenfeng: In the unit tests, `r3` is empty because the partitioned file associated with it (`p3`) starts after the beginning of the last indexed block. Previously, `r3` would have been nonempty as the partitioned file starts within the 0xFFFF margin of the last indexed block.

core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
```diff
@@ -482,17 +482,24 @@ object TabixIndexHelper extends GlowLogging {
...
-            val blockRangeStart = Math.max(file.start, minOverBlocks >> 16) // Shift 16 bits to get
-            // file offset of the bin from bgzipped virtual file offset.
-            val blockRangeEnd = Math.min(file.start + file.length, (maxOverBlocks >> 16) + 0xFFFF)
-            // 0xFFFF is the maximum possible length of an uncompressed bin.
-            if (blockRangeStart <= blockRangeEnd) {
-              Some((blockRangeStart, blockRangeEnd))
+            // Do not return any files without a BGZIP header
+            var fileContainsBlockStart = false
+            // Shift 16 bits to get file offset of the bin from bgzipped virtual file offset.
+            val (firstStart, firstLast) =
```
kianfar77, line 488: nit: Can we call this `firstEnd` instead of `firstLast`?

core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
```diff
@@ -482,17 +482,24 @@ object TabixIndexHelper extends GlowLogging {
...
+              (offsetList(0).getStartPosition >> 16, offsetList(0).getEndPosition >> 16)
+            val (minOverBlocks, maxOverBlocks) = offsetList.foldLeft((firstStart, firstLast)) {
+              case ((aStart, aEnd), o) =>
+                val bStart = o.getStartPosition >> 16
+                val bEnd = o.getEndPosition >> 16
+                if (!fileContainsBlockStart) {
+                  fileContainsBlockStart = (file.start <= bStart) && (file.start + file.length >= bStart)
+                }
+                (Math.min(aStart, bStart), Math.max(aEnd, bStart))
```
kianfar77, line 497: Do you mean `Math.max(aEnd, bEnd)`?

core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
```diff
@@ -482,17 +482,24 @@ object TabixIndexHelper extends GlowLogging {
...
-                case ((myStart, myEnd), e) =>
-                  (Math.min(myStart, e.getStartPosition), Math.max(myEnd, e.getEndPosition))
-              }
-            val blockRangeStart = Math.max(file.start, minOverBlocks >> 16) // Shift 16 bits to get
-            // file offset of the bin from bgzipped virtual file offset.
-            val blockRangeEnd = Math.min(file.start + file.length, (maxOverBlocks >> 16) + 0xFFFF)
-            // 0xFFFF is the maximum possible length of an uncompressed bin.
-            if (blockRangeStart <= blockRangeEnd) {
-              Some((blockRangeStart, blockRangeEnd))
+            // Do not return any files without a BGZIP header
```
kianfar77, line 485: nit: Change the comment to: `Flag used to avoid returning any files without a BGZIP header`

kianfar77, line 485: Can you please add the points you mention in the PR description regarding the behavior of the codec in reading the beginning and end of a block as well as the arbitrariness of spark made partitions as comments here? The will be important in understanding what is happening here.

kianfar77 (commented):
@karenfeng Great investigation into this! The fix looks correct to me. Just a few small comments.

core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
karenfeng: Whoops, good catch. Thanks!

core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
karenfeng: I don't know why I named it `firstLast` 🤦 . Thanks!

core/src/main/scala/io/projectglow/vcf/TabixIndexHelper.scala
karenfeng: I added it to the method description - can you confirm this includes the desired points?

kianfar77:
Beautiful!

kianfar77 (approved):

PR #318, opened 2020-12-23
Approximate Firth correction

karenfeng:

Implement approximate Firth correction for logistic regression.

## How is this patch tested?
- [x] Unit tests
- [x] Integration tests
- [ ] Manual tests

Includes end-to-end comparison versus regenie.


python/glow/gwas/log_reg.py
```diff
@@ -182,20 +197,48 @@ def _logistic_regression_inner(genotype_pdf: pd.DataFrame, log_reg_state: LogReg
     c, d: covariate
     '''
     X = np.column_stack(genotype_pdf[_VALUES_COLUMN_NAME].array)
+
+    # Based on Joelle's code
```
henrydavidge, line 201: Did you try this? I see that Joelle pointed out that in regenie they actually only use the result of this residualization for the firth correction.

python/glow/gwas/approx_firth_correction.py
```diff
@@ -0,0 +1,192 @@
...
+from dataclasses import dataclass
+from typeguard import typechecked
+from nptyping import Float, NDArray
+from scipy import stats
+
+
+@dataclass
+class LogLikelihood:
+    pi: NDArray[(Any,), Float]
+    G: NDArray[(Any, Any), Float] # diag(pi(1-pi))
```
henrydavidge, line 13: Comment for later: just store the diagonal as a vector.

python/glow/gwas/approx_firth_correction.py
```diff
@@ -0,0 +1,192 @@
...
+        X: NDArray[(Any, Any), Float],
+        y: NDArray[(Any,), Float],
+        offset: NDArray[(Any,), Float],
+        convergence_limit: float = 1e-5,
+        deviance_tolerance: float = 1e-6,
+        max_iter: int = 250,
+        max_step_size: int = 5,
+        max_half_steps: int = 25) -> Optional[FirthFit]:
+    '''
+    Firth’s bias-Reduced penalized-likelihood logistic regression.
```
henrydavidge, line 68: Comment that this is based on `logistf` and link to the original.

python/glow/gwas/approx_firth_correction.py
```diff
@@ -0,0 +1,192 @@
...
+
+    :param beta_init: Initial beta values
```
henrydavidge, line 70: Use the doc notation we use elsewhere.

python/glow/gwas/tests/test_approx_firth.py
```diff
@@ -0,0 +1,138 @@
...
+        X=np.column_stack([np.array(covariates).T, genotypes]),
+        y=np.array(phenotypes),
+        offset=np.array(offset),
+    )
+    golden_firth_beta = golden_firth_fit.head().beta
+    test_firth_beta = test_firth_fit.beta[-1]
+    assert np.allclose(golden_firth_beta, test_firth_beta)
+
+
+def test_full_firth(spark):
```
henrydavidge, line 61: For these tests, I'd actually prefer to just have literal golden values that can be cross referenced against R. I would prefer to not have a long term dependency on the SQL regression functions.

python/glow/gwas/tests/test_approx_firth.py
```diff
@@ -0,0 +1,138 @@
...
+    offset_trait1_df = _read_offset_df(test_data_dir + 'fit_bin_out_1.loco', 'Y1')
+    offset_trait2_df = _read_offset_df(test_data_dir + 'fit_bin_out_2.loco', 'Y2')
+    offset_df = pd.merge(offset_trait1_df, offset_trait2_df, left_index=True, right_index=True)
+
+    glowgr_df = lr.logistic_regression(genotype_df,
+                           phenotype_df,
+                           covariate_df,
+                           offset_df,
+                           correction=lr.correction_approx_firth,
+                           pvalue_threshold=0.9999, # correct all SNPs
```
henrydavidge, line 120: Should we not test the case where only some values are corrected?

python/glow/gwas/log_reg.py
```diff
@@ -182,20 +198,47 @@ def _logistic_regression_inner(genotype_pdf: pd.DataFrame, log_reg_state: LogReg
...
+                )
+                if approx_firth_snp_fit is not None:
+                    out_df.effect.iloc[correction_idx] = approx_firth_snp_fit.effect
+                    out_df.stderr.iloc[correction_idx] = approx_firth_snp_fit.stderr
+                    out_df.tvalue.iloc[correction_idx] = approx_firth_snp_fit.tvalue
+                    out_df.pvalue.iloc[correction_idx] = approx_firth_snp_fit.pvalue
+                else:
+                    print(f"Could not correct {out_df.iloc[correction_idx]}")
+        else:
+            raise ValueError(f"Only supported correction method is {correction_approx_firth}")
```
henrydavidge, line 242: This should be validated on the driver.

python/glow/gwas/log_reg.py
```diff
@@ -182,20 +198,47 @@ def _logistic_regression_inner(genotype_pdf: pd.DataFrame, log_reg_state: LogReg
...
+                    log_reg_state.approx_firth_state.logit_offset[:, pheno_idx],
+                    Y_mask[:, pheno_idx]
+                )
+                if approx_firth_snp_fit is not None:
+                    out_df.effect.iloc[correction_idx] = approx_firth_snp_fit.effect
+                    out_df.stderr.iloc[correction_idx] = approx_firth_snp_fit.stderr
+                    out_df.tvalue.iloc[correction_idx] = approx_firth_snp_fit.tvalue
+                    out_df.pvalue.iloc[correction_idx] = approx_firth_snp_fit.pvalue
+                else:
+                    print(f"Could not correct {out_df.iloc[correction_idx]}")
```
henrydavidge, line 240: This is not really helpful. We should put something in the result that indicates that the correction failed.

henrydavidge (commented):
Cool stuff! Had a bunch of comments.

python/glow/gwas/log_reg.py
```diff
@@ -77,12 +81,15 @@ def logistic_regression(
...
     genotype_df = gwas_fx._prepare_genotype_df(genotype_df, values_column, sql_type)
-    result_fields = [
-        # TODO: Probably want to put effect size and stderr here for approx-firth
+    base_result_fields = [
         StructField('tvalue', sql_type),
         StructField('pvalue', sql_type),
         StructField('phenotype', StringType())
     ]
+    if correction == correction_approx_firth:
+        result_fields = [StructField('effect', sql_type), StructField('stderr', sql_type)] + base_result_fields
```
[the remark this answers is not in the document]
karenfeng: I'm not sure if it's valuable to state the number of iterations, but it does make sense to clarify when the fit failed. I think it may make more sense to have a nested column with `beta`, `stderr`, and `converged` for Firth, which allows us to leave the uncorrected `tvalue` and `pvalue`.

python/glow/gwas/approx_firth.py
```diff
@@ -0,0 +1,178 @@
...
+        # Failed to converge
+        return None
+
+    return FirthFit(beta, log_likelihood)
+
+
+@typechecked
+def perform_null_firth_fit(y: NDArray[(Any, ), Float], C: NDArray[(Any, Any), Float],
+                           mask: NDArray[(Any, ), bool], offset: Optional[NDArray[(Any, ), Float]],
+                           fit_intercept: bool) -> NDArray[(Any, ), Float]:
```
henrydavidge, line 124: nit: perhaps it would be more clear to call `fit_intercept` `includes_intercept` here? `fit_intercept` in other functions means "add the intercept as a covariate," whereas here it means "the covariate matrix already includes an intercept".

python/glow/gwas/approx_firth.py
```diff
@@ -0,0 +1,178 @@
...
+    Returns:
+        None if the Firth fit did not converge.
+        Otherwise, offset vector with per-sample covariate effects for SNP fits.
+    '''
+
+    firth_offset = np.zeros(y.shape)
+    offset = offset[mask] if offset is not None else np.zeros(y.shape)
+    b0_null_fit = np.zeros(C.shape[1])
+    if fit_intercept:
+        b0_null_fit[-1] = (0.5 + y.sum()) / (mask.sum() + 1)
```
henrydavidge, line 138: Isn't the intercept the 0th column in the covariate matrix?

python/glow/gwas/approx_firth.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+# Skip typechecking for optimization
+def correct_approx_firth(x: NDArray[(Any, ), Float], y: NDArray[(Any, ), Float],
+                         firth_offset: NDArray[(Any, ), Float],
+                         y_mask: NDArray[(Any, ), bool]) -> Optional[FirthStatistics]:
+    '''
+    Calculate LRT statistics for a SNP using the approximate Firth method.
+
+    :return: None if the Firth fit did not converge. Otherwise, likelihood-ratio test statistics.
```
henrydavidge, line 155: Fix syntax

python/glow/gwas/log_reg.py
```diff
@@ -164,14 +185,17 @@ def _pdf_to_log_reg_state(pdf: pd.DataFrame, phenotypes: pd.Series, n_covar: int
...
 
 
 @typechecked
-def _create_log_reg_state(
-        spark: SparkSession, phenotype_df: pd.DataFrame, offset_df: pd.DataFrame,
-        sql_type: DataType, C: NDArray[(Any, Any),
-                                       Float]) -> Union[LogRegState, Dict[str, LogRegState]]:
+def _create_log_reg_state(spark: SparkSession, phenotype_df: pd.DataFrame, offset_df: pd.DataFrame,
+                          sql_type: DataType, C: NDArray[(Any, Any), Float], correction: str,
+                          fit_intercept: bool) -> Union[LogRegState, Dict[str, LogRegState]]:
```
henrydavidge, line 198: Same comment here about the parameter name for `fit_intercept`.

python/glow/gwas/log_reg.py
```diff
@@ -72,19 +74,27 @@ def logistic_regression(
         - ``tvalue``: The chi squared test statistic according to the score test or the correction method
         - ``pvalue``: P value estimated from the test statistic
         - ``phenotype``: The phenotype name as determiend by the column names of ``phenotype_df``
+        - ``effect``: The effect size (if approximate Firth correction was applied)
```
henrydavidge, line 77: nit: these should be in the same order as they appear in the output struct. Below it looks like we're putting these columns first.

python/glow/gwas/tests/test_approx_firth.py
```diff
@@ -0,0 +1,133 @@
...
+    offset_trait2_df = _read_offset_df(test_data_dir + 'fit_bin_out_2.loco', 'Y2')
+    offset_df = pd.merge(offset_trait1_df, offset_trait2_df, left_index=True, right_index=True)
+
+    glowgr_df = lr.logistic_regression(
+        genotype_df,
+        phenotype_df,
+        covariate_df,
+        offset_df,
+        correction=lr.correction_approx_firth,
+        pvalue_threshold=0.9999,  # correct all SNPs
```
henrydavidge, line 115: We should have a test where only some sites are corrected to verify that our checking logic is correct.

henrydavidge (commented):
Looking great! Comments remaining are relatively minor.

python/glow/gwas/approx_firth.py
karenfeng: Ah shoot, good catch.

karenfeng:
Thanks for the review @henrydavidge ! Can you take another look?

python/glow/gwas/approx_firth.py
```diff
@@ -0,0 +1,197 @@
...
+    beta = beta_init.copy()
+    model = sm.GLM(y, X, family=sm.families.Binomial(), offset=offset, missing='ignore')
+    log_likelihood = _calculate_log_likelihood(beta, model)
+
+    while n_iter < max_iter:
+        invI = np.linalg.pinv(log_likelihood.I)
+
+        # build hat matrix
+        rootG = np.sqrt(log_likelihood.pi * (1 - log_likelihood.pi))
+        rootG_X = np.expand_dims(rootG, 1) * X  # equivalent to np.diagflat(pi * (1 - pi)) @ X
```
henrydavidge, line 82: nits: I think it's more idiomatic (and common in our codebase) to use `rootG[:, None]`. Also, the comment doesn't include the `sqrt`.

python/glow/gwas/approx_firth.py
```diff
@@ -0,0 +1,197 @@
...
+
+    effect = firth_fit.beta.item()
+    # Likelihood-ratio test
+    null_model = sm.GLM(masked_y,
+                        masked_X,
+                        family=sm.families.Binomial(),
+                        offset=masked_offset,
+                        missing='ignore')
+    null_deviance = _calculate_log_likelihood(beta_init, null_model).deviance
+    tvalue = -1 * (firth_fit.log_likelihood.deviance - null_deviance)
```
henrydavidge, line 193: nit: maybe just write as `null_deviance - firth_fit.log_likelihood.deviance`

python/glow/gwas/log_reg.py
```diff
@@ -69,6 +71,10 @@ def logistic_regression(
         A Spark DataFrame that contains
 
         - All columns from ``genotype_df`` except the ``values_column`` and the ``genotypes`` column if one exists
+        - ``effect``: The effect size (if approximate Firth correction was applied)
+        - ``stderror``: Standard error of the effect size (if approximate Firth correction was applied)
+        - ``correction``: The correction method used (if the correction test method is not ``none``).
```
henrydavidge, line 76: Hm, I'm not sure about this semantic. It should be straightforward to know if there was a fitting error. Maybe it would be more user friendly to have a `fit_status` field? Where if positive the status indicates the number of iterations to fit, `-1` means no convergence, `-2` means exploded. Or even a boolean that indicates if the correction succeeded.

python/glow/gwas/log_reg.py
```diff
@@ -242,20 +282,44 @@ def _logistic_regression_inner(genotype_pdf: pd.DataFrame, log_reg_state: LogReg
...
-        ()
-
     del genotype_pdf[_VALUES_COLUMN_NAME]
     out_df = pd.concat([genotype_pdf] * log_reg_state.Y_res.shape[1])
     out_df['tvalue'] = list(np.ravel(t_values))
     out_df['pvalue'] = list(np.ravel(p_values))
     out_df['phenotype'] = phenotype_names.repeat(genotype_pdf.shape[0]).tolist()
+
+    if correction != correction_none:
+        out_df['correction_method'] = 'none'
```
henrydavidge, line 304: I'm still not convinced we need this field. Seems like it would be sufficient to just have the `correction_succeeded` column -- `True`, `False`, or `null` for no correction.

henrydavidge (approved):
Just one more comment on how we use the `correctioN_status` and `correction_method` columns. You can make the final call and then merge.

PR #319, opened 2020-12-29
Parallelize null model fit for logistic regression

henrydavidge:
Fitting all the null models and preparing the broadcast data could take a looooong time with many phenotypes and loco offsets. Let's use Spark!

I also sped up the `reshape_for_gwas` function.

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -22,7 +22,8 @@ def _spark_builder():
     return SparkSession.builder \
         .master("local[2]") \
         .config("spark.hadoop.io.compression.codecs", "io.projectglow.sql.util.BGZFCodec") \
-        .config("spark.ui.enabled", "false")
+        .config("spark.ui.enabled", "false") \
+        .config("spark.sql.execution.arrow.pyspark.enabled", "true")
 
 # sbt guarantees that this environment variable is set for python tests
 SPARK_VERSION = os.environ['SPARK_VERSION']
--- a/python/glow/gwas/log_reg.py
+++ b/python/glow/gwas/log_reg.py
@@ -1,8 +1,8 @@
-from pyspark.sql.types import StringType, StructField
-from typing import Any, Optional
+from pyspark.sql.types import ArrayType, StringType, StructField, DataType, StructType
+from typing import Any, Optional, Dict, Union
 import pandas as pd
 import numpy as np
-from pyspark.sql import DataFrame
+from pyspark.sql import DataFrame, SparkSession
 import statsmodels.api as sm
 from dataclasses import dataclass
 from typeguard import typechecked
@@ -10,7 +10,8 @@ from nptyping import Float, NDArray
 from scipy import stats
 import opt_einsum as oe
 from . import functions as gwas_fx
-from .functions import _VALUES_COLUMN_NAME
+from .functions import _OffsetType, _VALUES_COLUMN_NAME
+from ..wgr.functions import reshape_for_gwas
 
 __all__ = ['logistic_regression']
 
@@ -73,7 +74,8 @@ def logistic_regression(
         - ``phenotype``: The phenotype name as determiend by the column names of ``phenotype_df``
     '''
 
-    gwas_fx._check_spark_version(genotype_df.sql_ctx.sparkSession)
+    spark = genotype_df.sql_ctx.sparkSession
+    gwas_fx._check_spark_version(spark)
     gwas_fx._validate_covariates_and_phenotypes(covariate_df, phenotype_df, is_binary=True)
     sql_type = gwas_fx._regression_sql_type(dt)
     genotype_df = gwas_fx._prepare_genotype_df(genotype_df, values_column, sql_type)
@@ -92,9 +94,7 @@ def logistic_regression(
     Y_mask = ~(np.isnan(Y))
     np.nan_to_num(Y, copy=False)
 
-    state = gwas_fx._loco_make_state(
-        Y, phenotype_df, offset_df,
-        lambda y, pdf, odf: _create_log_reg_state(y, pdf, odf, C, Y_mask))
+    state = _create_log_reg_state(spark, phenotype_df, offset_df, sql_type, C)
 
     phenotype_names = phenotype_df.columns.to_series().astype('str')
 
@@ -106,18 +106,18 @@ def logistic_regression(
     return genotype_df.mapInPandas(map_func, result_struct)
 
 
-@typechecked
-def _logistic_null_model_predictions(
-    y: NDArray[Any, Float],
-    X: NDArray[Any, Float],
-    y_mask: NDArray[Any, bool],
-    # nptying can't handle optional NDArrays, so don't verify the type for offset
-    offset
-) -> NDArray[Any, Float]:
+@dataclass
+class LogRegState:
+    inv_CtGammaC: NDArray[(Any, Any), Float]
+    gamma: NDArray[(Any, Any), Float]
+    Y_res: NDArray[(Any, Any), Float]
+
+
+def fit_null_model(y, X, mask, offset):
     if offset is not None:
-        offset = offset[y_mask]
-    model = sm.GLM(y[y_mask],
-                   X[y_mask],
+        offset = offset[mask]
+    model = sm.GLM(y[mask],
+                   X[mask, :],
                    family=sm.families.Binomial(),
                    offset=offset,
                    missing='ignore')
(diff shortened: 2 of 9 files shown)


conftest.py
```diff
@@ -22,7 +22,8 @@ def _spark_builder():
     return SparkSession.builder \
         .master("local[2]") \
         .config("spark.hadoop.io.compression.codecs", "io.projectglow.sql.util.BGZFCodec") \
-        .config("spark.ui.enabled", "false")
+        .config("spark.ui.enabled", "false") \
+        .config("spark.sql.execution.arrow.pyspark.enabled", "true")
```
henrydavidge, line 26: Most environments (including Databricks) should have this enabled by now. It allows us to avoid converting numpy arrays to python lists when converting Spark <=> Pandas DataFrames.

python/glow/wgr/functions.py
```diff
@@ -179,11 +179,17 @@ def reshape_for_gwas(spark: SparkSession, label_df: pd.DataFrame) -> DataFrame:
...
+        contigs = label_df.index.get_level_values(1).unique()
+        transposed_df = pd.concat([transpose_one(contig) for contig in contigs],
+                                  keys=contigs,
+                                  names=['contig', 'label'])
+        column_names = ['contigName', 'label', 'values']
     else:
         raise ValueError('label_df must be indexed by sample id or by (sample id, contig name)')
 
-    transposed_df['values_array'] = transposed_df.to_numpy().tolist()
+    transposed_df['values_array'] = list(transposed_df.to_numpy())
```
henrydavidge, line 194: We can use this faster conversion when arrow is enabled.

python/glow/wgr/functions.py
```diff
@@ -179,11 +179,17 @@ def reshape_for_gwas(spark: SparkSession, label_df: pd.DataFrame) -> DataFrame:
...
         # ordering in case it's not sorted
-        ordered_cols = pd.unique(label_df.index.get_level_values(0))
-        transposed_df = label_df.T.stack()[ordered_cols]
-        column_names = ['label', 'contigName', 'values']
+        def transpose_one(contig):
+            transposed = label_df.xs(contig, level=1).T
+            return transposed
+
+        contigs = label_df.index.get_level_values(1).unique()
+        transposed_df = pd.concat([transpose_one(contig) for contig in contigs],
```
henrydavidge, line 187: It turns out that `stack` is crazy slow. This method is >10x faster.

python/glow/gwas/log_reg.py
```diff
@@ -163,7 +220,7 @@ def _logistic_residualize(X: NDArray[(Any, Any), Float], C: NDArray[(Any, Any),
     Residualize the genotype vectors given the null model predictions.
     X_res = X - C(C.T gamma C)^-1 C.T gamma X
     '''
-    X_hat = gwas_fx._einsum('ic,pcd,ds,ps,sg,sp->igp', C, inv_CtGammaC, C.T, gamma, X, Y_mask)
+    X_hat = gwas_fx._einsum('ic,pcd,ds,sp,sg,sp->igp', C, inv_CtGammaC, C.T, gamma, X, Y_mask)
```
karenfeng, line 223: What happened here?

python/glow/gwas/tests/test_lin_reg.py
```diff
@@ -19,7 +19,7 @@ def run_linear_regression(genotype_df, phenotype_df, covariate_df, fit_intercept
     Y[~Y_mask] = 0
     Q = np.linalg.qr(C)[0]
     Y = lr._residualize_in_place(Y, Q)
-    Y_state = lr._create_YState(Y, phenotype_df, None, Y_mask, np.float64)
+    Y_state = lr._create_YState(Y, phenotype_df, pd.DataFrame({}), Y_mask, np.float64)
```
karenfeng, line 22: Why did we change this?

python/glow/gwas/tests/test_log_reg.py
```diff
@@ -13,7 +13,14 @@ def run_score_test(genotype_df, phenotype_df, covariate_df, fit_intercept=True):
     Y = phenotype_df.to_numpy(copy=True)
     Y_mask = ~np.isnan(Y)
     Y[~Y_mask] = 0
-    state = lr._create_log_reg_state(Y, pd.DataFrame(), None, C, Y_mask)
+    state_rows = [
+        lr._prepare_one_phenotype(C, pd.Series({
+            'label': c,
+            'values': phenotype_df[c]
+        })) for c in phenotype_df
```
karenfeng, line 20: Nit: can we use `p` instead of `c` to represent the phenotypes?

python/glow/gwas/tests/test_log_reg.py
```diff
@@ -13,7 +13,14 @@ def run_score_test(genotype_df, phenotype_df, covariate_df, fit_intercept=True):
...
+    ]
+    phenotype_names = phenotype_df.columns.to_series().astype('str')
+    state = lr._pdf_to_log_reg_state(pd.DataFrame(state_rows), phenotype_names, C.shape[1])
```
karenfeng, line 23: Why didn't we use `_create_log_reg_state ` here?

karenfeng (commented):
Took a first pass. Thanks for doing this!

conftest.py
henrydavidge: Our unit tests for reshaping are pretty short. I don't think it would affect test time elsewhere.

python/glow/gwas/log_reg.py
```diff
@@ -106,18 +106,18 @@ def map_func(pdf_iterator):
...
-    offset
-) -> NDArray[Any, Float]:
+@dataclass
+class LogRegState:
+    inv_CtGammaC: NDArray[(Any, Any), Float]
+    gamma: NDArray[(Any, Any), Float]
+    Y_res: NDArray[(Any, Any), Float]
+
+
+def _logistic_null_model_predictions(y, X, mask, offset):
```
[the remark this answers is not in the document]
henrydavidge: It's usually a good idea to keep functions called within UDFs as trim as possible since they can be called many times. I don't believe typechecking is very expensive, but it can add up.

python/glow/gwas/log_reg.py
```diff
@@ -126,34 +126,91 @@ def _logistic_null_model_predictions(
...
+    '''
+    y = row['values']
+    mask = ~np.isnan(y)
+    y_pred = _logistic_null_model_predictions(y, C, mask, row.get('offset'))
+    y_res = np.nan_to_num(y - y_pred)
+    gamma = y_pred * (1 - y_pred)
+    CtGammaC = C.T @ (gamma[:, None] * C)
+    inv_CtGammaC = np.linalg.inv(CtGammaC)
+    row.label = str(row.label)
+    row.drop(['values', 'offset'], inplace=True, errors='ignore')
```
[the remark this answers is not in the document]
henrydavidge: You mean to remove the `inplace`? I don't think it would make a significant perf difference, no. The whole null model fitting process doesn't actually matter much for perf since it is MUCH faster (like tens of seconds vs hours) than the association scan. It was only a problem when we were fitting all the models serially.

The main reason I used `row.drop` here was for simplicity -- otherwise we need separate cases for the loco vs non-loco paths.

python/glow/gwas/log_reg.py
```diff
@@ -126,34 +126,91 @@ def _logistic_null_model_predictions(
...
+
+
+@typechecked
+def _pdf_to_log_reg_state(pdf: pd.DataFrame, phenotypes: pd.Series, n_covar: int) -> LogRegState:
+    '''
+    Converts a Pandas DataFrame with the contents of a LogRegState object
+    into a more convenient form.
+    '''
+    sorted_pdf = pdf.set_index('label').reindex(phenotypes, axis=0)
+    inv_CtGammaC = np.row_stack(sorted_pdf['inv_CtGammaC'].array).reshape(
```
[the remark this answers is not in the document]
henrydavidge: Ah, the type hint on `LogRegState` was not correct. It's a 3D ndarray, as demonstrated in the einsum contractions. It's basically an `n_covar x n_covar` matrix per phenotype.

python/glow/gwas/log_reg.py
henrydavidge: The way that the null model fitting code was written before, it was more convenient for `gamma` to have the shape `n_phenotype x n_sample`. However, on second reading I found that confusing since we typically have one sample per row in the gwas functions. So I transposed `gamma` in the new null fit code.

python/glow/gwas/tests/test_log_reg.py
henrydavidge: I wanted to be able to run a basic test without using Spark. It's much easier to experiment and debug when Spark isn't involved.

python/glow/gwas/log_reg.py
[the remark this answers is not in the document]
henrydavidge: I fixed the type annotation. Tried adding a `@typechecked` decorator to the dataclasses, but it doesn't seem to do anything.

karenfeng (approved):
LGTM overall, thanks for doing this! I left some small requests for clarification; feel free to merge once addressed.

python/glow/gwas/tests/test_lin_reg.py
henrydavidge: With the new changes it was easier to take an empty dataframe since that matches the overall `linear_regression` function. It's private, so not a big deal to change.

PR #324, opened 2021-01-14
GWAS doc improvements

karenfeng:

Improves GWAS documentation to reflect the new logistic regression function, and replace outdated references to the SQL functions.

I also had to move `_get_contigs_from_loco_df` due to circular dependencies.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

Ran notebook and visualized changes with `make livehtml`.


python/glow/gwas/log_reg.py
```diff
@@ -38,15 +38,16 @@ def logistic_regression(genotype_df: DataFrame,
     On the driver node, we fit a logistic regression model based on the covariates for each
```
kianfar77, line 38: Add "Implementation details" to before this paragraph like what we have for lin_reg

docs/source/tertiary/regression-tests.rst
```diff
@@ -110,211 +110,156 @@ For complete parameter usage information, check out the API reference for :func:
...
-      lit(covariates.to_numpy()),
-      'Firth',
-      lit(offset)
-    ))
+  assert_rows_equal(
+    log_reg_df.filter("contigName = 22 and start = 16050114 and phenotype = 'Binary_Trait_1'").head(),
+    expected_uncorrected_row
   )
 
+The logistic regression function accepts GloWGR phenotypic predictions (either global or per chromosome) as an offset.
```
kianfar77, line 197: This sentence reads along with your tip below is a bit repetitive. I suggest we give this offset thing their own subsection in this page both under linear and under logistic regression. I think it is important enough to get its own subsections.

kianfar77 (commented):
Looks great. Just a couple of minor suggestions.

python/glow/gwas/log_reg.py
```diff
@@ -74,10 +75,10 @@ def logistic_regression(genotype_df: DataFrame,
         - ``effect``: The effect size (if approximate Firth correction was applied)
```
kianfar77, line 75: This is unrelated to this PR but occurred to me while working on WGR. If @henrydavidge and you agree, I think it is better we change the `fit_intercept` argument to `add_intercept`. I know that scikit calls it `fit_intercept` but I think it is a bit misleading because the cov matrix can have the intercept already in it and therefore fitting is actually happening in that case anyway. This should be about whether the user want us to add that column to covariates or not.

python/glow/gwas/__init__.py
```diff
@@ -1,2 +1,4 @@
 from .lin_reg import *
-__all__ = ['linear_regression']
\ No newline at end of file
+from .log_reg import *
+
+__all__ = lin_reg.__all__ + log_reg.__all__
```
henrydavidge, line 4: Huh, I'm surprised this works. I didn't think that `import * from foo` imports `foo` itself.

docs/source/_static/zzz_GENERATED_NOTEBOOK_SOURCE/tertiary/gwas.py
```diff
@@ -281,18 +276,16 @@ class Covariates:
 sample_info = get_sample_info(vcf_view, sample_metadata)
 sample_count = sample_info.count()
 mlflow.log_param("number of samples", sample_count)
-pcs_indexed = pcs_df.coalesce(1).withColumn("index", monotonically_increasing_id())
+pcs_indexed = pcs_df.coalesce(1).withColumn("index", fx.monotonically_increasing_id())
```
henrydavidge, line 279: It might make more sense to put the PCs into a CSV and just read that with pandas. I think that'll be a more common input format anyway.

karenfeng: spark.read.format("delta").load(principal_components_path)
covariate_df = pd.DataFrame(covariates.head().covariates.toArray())

karenfeng: This actually does not touch the datasource, and is only used for visualization. I'll switch to CSV in between though.

henrydavidge (approved):
LGTM

kianfar77 (approved):
Thanks Karen

PR #326, opened 2021-01-19
Improve register

henrydavidge:
While running the example notebooks on a Databricks cluster, I noticed that some functions can fail with confusing analysis errors if the cluster has run a query (including showing tables) before `glow.register` was called. To improve the default behavior, I added a `new_session` parameter to `register` so that this error can't happen by default.

In addition, I updated the register method to decrease the columnar parquet reader's batch size by default, which is typically necessary for large cohorts. This matches what we do in the Databricks genomics runtime.

I will update the notebooks in a forthcoming notebook cleanup PR.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -21,7 +21,6 @@ import os
 def _spark_builder():
     return SparkSession.builder \
         .master("local[2]") \
-        .config("spark.hadoop.io.compression.codecs", "io.projectglow.sql.util.BGZFCodec") \
         .config("spark.ui.enabled", "false") \
         .config("spark.sql.execution.arrow.pyspark.enabled", "true")
 
--- a/core/src/test/scala/io/projectglow/sql/GlowBaseTest.scala
+++ b/core/src/test/scala/io/projectglow/sql/GlowBaseTest.scala
@@ -18,7 +18,7 @@ package io.projectglow.sql
 
 import htsjdk.samtools.util.Log
 import org.apache.spark.sql.SparkSession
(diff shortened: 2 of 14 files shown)


conftest.py
```diff
@@ -21,7 +21,6 @@
 def _spark_builder():
     return SparkSession.builder \
         .master("local[2]") \
-        .config("spark.hadoop.io.compression.codecs", "io.projectglow.sql.util.BGZFCodec") \
```
henrydavidge, line 24: We set this config in the VCF reader / writer, so we don't need it here.

core/src/test/scala/io/projectglow/sql/GlowBaseTest.scala
```diff
@@ -36,18 +35,11 @@ abstract class GlowBaseTest
     with TestUtils
     with JenkinsTestPatience {
 
-  override protected def sparkConf: SparkConf = {
-    super
-      .sparkConf
-      .set("spark.hadoop.io.compression.codecs", classOf[BGZFCodec].getCanonicalName)
-      .set(GlowConf.FAST_VCF_READER_ENABLED.key, "false") // TODO(hhd): Enable the fast reader once we're confident
```
henrydavidge, line 43: Oops, I should have removed this a while ago.

kianfar77 (approved):
LGTM. Thanks

PR #328, opened 2021-01-22
Prepare phenotypes

karenfeng:

In regenie, the phenotypes are transformed before being passed to the regression function:
- [Mean-centered](https://github.com/rgcgithub/regenie/blob/a8f110dc87b8566e280b8a35bc30a86670bbd2d1/src/Pheno.cpp#L226)
- [Residualized with covariates](https://github.com/rgcgithub/regenie/blob/a8f110dc87b8566e280b8a35bc30a86670bbd2d1/src/Pheno.cpp#L399)
- [Scaled with an unbiased estimate of the standard deviation](https://github.com/rgcgithub/regenie/blob/a8f110dc87b8566e280b8a35bc30a86670bbd2d1/src/Pheno.cpp#L591)

[The offsets](https://github.com/rgcgithub/regenie/blob/a8f110dc87b8566e280b8a35bc30a86670bbd2d1/src/Data.cpp#L2418) are already on the scale of the prepared phenotypes.

Before the results are returned, they are [adjusted to reflect the original scale](https://github.com/rgcgithub/regenie/blob/a8f110dc87b8566e280b8a35bc30a86670bbd2d1/src/Data.cpp#L1485).

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

WIP: I also need to create an end-to-end test versus regenie.


python/glow/gwas/lin_reg.py
```diff
@@ -63,8 +63,9 @@ def linear_regression(genotype_df: DataFrame,
         phenotype_df : Pandas DataFrame containing phenotypic data
         covariate_df : An optional Pandas DataFrame containing covariates
         offset_df : An optional Pandas DataFrame containing the phenotype offset. The actual phenotype used
-                    for linear regression is ``phenotype_df`` minus the appropriate offset. The ``offset_df`` may
-                    have one or two levels of indexing. If one level, the index should be the same as the ``phenotype_df``.
+                    for linear regression is the mean-centered, residualized and scaled ``phenotype_df`` minus the
```
henrydavidge, line 66: nit: clarify that this matches what comes out of GloWGR or regenie

python/glow/gwas/lin_reg.py
```diff
@@ -114,16 +115,20 @@ def linear_regression(genotype_df: DataFrame,
     Q = np.linalg.qr(C)[0]
     Y = phenotype_df.to_numpy(dt, copy=True)
     Y_mask = (~np.isnan(Y)).astype(dt)
-    np.nan_to_num(Y, copy=False)
-    Y = gwas_fx._residualize_in_place(Y, Q)
+    Y = np.nan_to_num(Y, copy=False)
+    Y -= Y.mean(axis=0)  # Mean-center
+    Y = gwas_fx._residualize_in_place(Y, Q) * Y_mask  # Residualize
+    Y_scale = np.sqrt(np.sum(Y**2, axis=0) / (Y_mask.sum(axis=0) - Q.shape[1]))
+    Y /= Y_scale[np.newaxis, :]  # Scale
```
henrydavidge, line 122: nit: I think we typically use `None` instead of `np.newaxis`. They do the same thing.

python/glow/gwas/lin_reg.py
```diff
@@ -157,15 +162,17 @@ def _create_YState(Y: NDArray[(Any, Any), Float], phenotype_df: pd.DataFrame,
 def _create_one_YState(Y: NDArray[(Any, Any), Float], phenotype_df: pd.DataFrame,
                        offset_df: pd.DataFrame, Y_mask: NDArray[(Any, Any), Float], dt) -> YState:
     if not offset_df.empty:
-        Y = (pd.DataFrame(Y, phenotype_df.index, phenotype_df.columns) - offset_df).to_numpy(dt)
+        base_Y = pd.DataFrame(Y, phenotype_df.index, phenotype_df.columns)
+        Y = (base_Y - offset_df).reindex(phenotype_df.index).to_numpy(dt)
```
henrydavidge, line 166: Please add a comment about why we reindex

python/glow/gwas/lin_reg.py
```diff
@@ -189,11 +199,13 @@ def _linear_regression_inner(genotype_pdf: pd.DataFrame, Y_state: YState,
     pvalues = 2 * stats.distributions.t.sf(np.abs(T), dof)
 
     del genotype_pdf[_VALUES_COLUMN_NAME]
+    num_genotypes = genotype_pdf.shape[0]
     out_df = pd.concat([genotype_pdf] * Y_state.Y.shape[1])
-    out_df['effect'] = list(np.ravel(betas))
-    out_df['stderror'] = list(np.ravel(standard_error))
+    Y_scale_mat = np.expand_dims(Y_scale, axis=1)
```
henrydavidge, line 204: This is the same as `[:, None]`, right? If so please use that for consistency.

python/glow/gwas/tests/test_lin_reg.py
```diff
@@ -16,13 +17,16 @@ def run_linear_regression(genotype_df, phenotype_df, covariate_df, add_intercept
     Y = phenotype_df.to_numpy('float64', copy=True)
     Y_mask = ~np.isnan(Y)
     Y[~Y_mask] = 0
+    Y -= Y.mean(axis=0)
     Q = np.linalg.qr(C)[0]
-    Y = gwas_fx._residualize_in_place(Y, Q)
+    Y = gwas_fx._residualize_in_place(Y, Q) * Y_mask
+    Y_std_dev = np.sqrt(np.sum(Y**2, axis=0) / (Y_mask.sum(axis=0) - Q.shape[1]))
```
henrydavidge, line 23: It actually might make sense to not do the scaling here since we have no offset. Then we have a simpler baseline to test against. What do you think?

python/glow/gwas/tests/test_lin_reg.py
```diff
@@ -343,9 +359,34 @@ def test_multi_offset_with_missing(spark, rg):
...
+    extra_cols = pd.DataFrame({'contigName': contigs * 6})
+    with pytest.raises(ValueError):
+        run_linear_regression_spark(spark,
+                                    genotype_df,
+                                    phenotype_df,
+                                    covariate_df,
+                                    offset_df=offset_df,
+                                    extra_cols=extra_cols)
+
+
```
henrydavidge, line 389: Can we have a test explicitly for the offset ordering thing?

henrydavidge (approved):
Woohoo!

PR #329, opened 2021-01-26
Make mean_substitute faster

henrydavidge:

Turns out that `mean_substitute` was accidentally `O(n^2)` in the worst case. We recomputed the mean for each missing element.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)

### DIFF
--- a/core/src/main/scala/io/projectglow/sql/dsl/package.scala
+++ b/core/src/main/scala/io/projectglow/sql/dsl/package.scala
@@ -24,15 +24,25 @@ package object dsl {
 
   trait ImplicitOperators {
     def expr: Expression
+
+    // Ensure that lambda variables have unique names for nested functions
+    private var lambdaCounter = 0
+    private def nextLambdaName(): String = {
+      lambdaCounter += 1
+      lambdaCounter.toString
+    }
+
     private def makeLambdaFunction(f: Expression => Expression): LambdaFunction = {
-      val x = UnresolvedNamedLambdaVariable(Seq("x"))
+      val x = UnresolvedNamedLambdaVariable(Seq(nextLambdaName()))
       LambdaFunction(f(x), Seq(x))
     }
+
     private def makeLambdaFunction(f: (Expression, Expression) => Expression): LambdaFunction = {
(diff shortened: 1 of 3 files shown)


core/src/test/scala/io/projectglow/tertiary/VariantQcExprsSuite.scala
```diff
@@ -352,6 +352,18 @@ class VariantQcExprsSuite extends GlowBaseTest {
     assert(test == Seq(1.5, 1.5, 0.0, 1.0, 2.0, 3.0))
   }
 
+  test("mean substitute big array") {
```
henrydavidge, line 355: This test timed out before

core/src/main/scala/io/projectglow/sql/expressions/MeanSubstitute.scala
```diff
@@ -107,7 +98,10 @@ case class MeanSubstitute(array: Expression, missingValue: Expression)
         s"Missing value must be of numeric type; provided type is ${missingValue.dataType}.")
     }
 
-    // Replace missing values with the provided strategy
-    array.arrayTransform(substituteWithMean(_))
+    array.aggregate(
```
karenfeng, line 101: Can you leave a clarifying comment here on why we need this for performance?

core/src/main/scala/io/projectglow/sql/dsl/package.scala
```diff
@@ -24,15 +24,25 @@ package object dsl {
 
   trait ImplicitOperators {
     def expr: Expression
+
+    // Ensure that lambda variables have unique names for nested functions
+    private var lambdaCounter = 0
+    private def nextLambdaName(): String = {
```
karenfeng, line 30: How does this change things? Do we have tests?

henrydavidge: We couldn't have nested higher order functions before. So the new implementation of `mean_substitute` failed analysis. I don't think we need a particular test here.

karenfeng (approved):
LGTM, thanks!

PR #341, opened 2021-02-03
Perform Hail tests separately during release

karenfeng:

Hail tests require additional configuration to our normal tests. In addition, it can mess with the current Python setup (by relying on old pandas versions, etc.). To avoid Hail artificially breaking our tests, I broke out the tests during the release process so we can configure them individually.

## How is this patch tested?
- [ ] Unit tests
- [x] Integration tests
- [ ] Manual tests

Discovered spurious failures as a result of running a release; these changes un-broke the release.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -452,8 +452,8 @@ Seq(
   commitStableVersion,
   tagRelease
 ) ++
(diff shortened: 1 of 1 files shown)


kianfar77 (approved):
Thanks Karen

PR #366, opened 2021-04-17
Remove typecheck for numpy arrays 

kianfar77:

`@typechecked` directive from typeguard package does not support numpy arrays. It has an unpredictable behavior. For a safer behavior, this PR deactivates type checking  for functions that have numpy array arguments. To keep type checking for the main entry points (`RidgeReduction`, `LogisticRidgeRegression` and `RidgeRegression`), I changed the `alphas` parameter type for them to  `List[float]`.

Also changed `alpha_label_coef.label` column to `alpha_label_coef_label` to avoid spark problem with dot in name

## How is this patch tested?
- [ x ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


henrydavidge:
Interesting, we use numpy type annotations extensively. We don't actually enforce the types anywhere else?

kianfar77:
> Interesting, we use numpy type annotations extensively. We don't actually enforce the types anywhere else?

It shows an unpredictable behavior. For some functions it works. For some does not. It has caused problems for customers. I now deactivated it for more functions. For our entry point classes, I changed the a`alphas` type to keep the typecheck.

karenfeng (approved):

karenfeng:
Thanks @kianfar77! Let's make sure to cut a new release for this.

PR #376, opened 2021-05-27
update introduction and architecture diagram

williambrandler:

Added more details to the introduction and updated the architecture diagram

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


williambrandler:
@henrydavidge updated the arch diagram and intro, what do you think?

karenfeng:
Thanks for the PR @williambrandler! I'm a little confused by the diagram. Can you clarify a couple points?
- Are we calculating allele frequency from VEP/variant annotation Delta table, or from the variant data before VEP?
- Are gene/region annotations relevant for gene/variant association? Is this for region-based association tests?
- Do the phenotypes/covariates go directly into a Delta table containing associations, or should they go through Glow first?

williambrandler:
thanks, yeah I tried to put too much in the diagram! 
Good catch on the phenos. Incorporated your suggestions and simplified

karenfeng:
Thanks for addressing my comments! LGTM now, I'll merge.

PR #377, opened 2021-05-28
Hail related enhancements

bcajes:
Extending linreg gwas functionality to support optional output of n, sum_x, and y_transpose_x .  Similar to what is output in [hail linreg](https://github.com/hail-is/hail/blob/8579a3f62529ca0e2787fdb84c5ac32b7b39d2f2/hail/python/hail/methods/statgen.py#L491), and useful for a more detailed analysis. I've also added a workaround/fix for a NPE bug when deriving genotype_state from a hail matrix with null genotype calls.  Added unit tests and verified new stats and NPE fix on production data.

### DIFF
--- a/core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
+++ b/core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
@@ -41,25 +41,20 @@ object VariantUtilExprs {
       val calls = genotypes
         .getStruct(i, genotypesSize)
         .getArray(genotypesIdx)
-      if (calls == null) {
-        output.setInt(i, -1) // missing
-        i += 1
-      } else {
-        var sum = 0
-        var j = 0
-        while (j < calls.numElements() && sum >= 0) {
-          if (calls.getInt(j) >= 0) {
-            sum += calls.getInt(j)
-          } else {
-            sum = -1 // missing
-          }
-          j += 1
+      var sum = 0
+      var j = 0
+      while (j < calls.numElements() && sum >= 0) {
+        if (calls.getInt(j) >= 0) {
+          sum += calls.getInt(j)
+        } else {
+          sum = -1 // missing
         }
-
-        output.setInt(i, if (j == 0) -1 else sum)
-        i += 1
+        j += 1
       }
+      output.setInt(i, if (j == 0) -1 else sum)
+      i += 1
     }
+
     output
   }
 
--- a/core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
+++ b/core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
@@ -41,20 +41,24 @@ object VariantUtilExprs {
       val calls = genotypes
         .getStruct(i, genotypesSize)
         .getArray(genotypesIdx)
-      var sum = 0
-      var j = 0
-      while (j < calls.numElements() && sum >= 0) {
-        if (calls.getInt(j) >= 0) {
-          sum += calls.getInt(j)
-        } else {
-          sum = -1 // missing
+      if (calls == null) {
+        output.setInt(i, -1) // missing
+        i += 1
+      } else {
+        var sum = 0
+        var j = 0
+        while (j < calls.numElements() && sum >= 0) {
+          if (calls.getInt(j) >= 0) {
+            sum += calls.getInt(j)
+          } else {
+            sum = -1 // missing
+          }
+          j += 1
         }
-        j += 1
+        output.setInt(i, if (j == 0) -1 else sum)
+        i += 1
       }
-      output.setInt(i, if (j == 0) -1 else sum)
-      i += 1
     }
-
     output
   }
 
--- a/core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
+++ b/core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
@@ -55,6 +55,7 @@ object VariantUtilExprs {
           }
           j += 1
         }
+
         output.setInt(i, if (j == 0) -1 else sum)
         i += 1
       }
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,4 +1,25 @@
-The Glow documentation is published with [Read the Docs](https://readthedocs.org/).
+##### The Glow documentation is published with [Read the Docs](https://readthedocs.org/).
 
-To build the docs locally, activate the dev conda environment in `docs/source/environment.yml` and then
-run `make livehtml` from this directory.
+To build the docs locally,
+first create the conda environment:
+
+`conda env create -f source/environment.yml`
+
+activate the glow docs dev conda [environment](source/environment.yml):
+
+`conda activate glow-docs`
+
+build the docs:
+
+`make livehtml`
+
+connect to the local server via your browser at:
+
+`http://127.0.0.1:8000`
+
+when a change is detected in the docs, 
+the documentation is rebuilt and any open browser windows are reloaded automatically.
+
+deactivate the environment after you are done:
+
+`conda deactivate`
--- a/docs/source/etl/variant-data.rst
+++ b/docs/source/etl/variant-data.rst
@@ -72,7 +72,7 @@ You can control the behavior of the VCF reader with a few parameters. All parame
 
    Glow includes a VCF reader that uses `htsjdk <https://github.com/samtools/htsjdk>`_ for initial parsing as well as a reader that parses VCF lines to Spark rows directly.
 
-   As of release 0.7.0, the direct reader is enabled by default. To use the htsjdk based reader, set the Spark config ``io.projectglow.vcf.fastReaderEnabled`` to ``false``.
+   As of release 1.0.0, the direct reader is enabled by default. To use the htsjdk based reader, set the Spark config ``io.projectglow.vcf.fastReaderEnabled`` to ``false``.
 
 
 .. important:: The VCF reader uses the 0-start, half-open (zero-based) coordinate system. This means
--- a/python/glow/gwas/lin_reg.py
+++ b/python/glow/gwas/lin_reg.py
@@ -79,6 +79,8 @@ def linear_regression(genotype_df: DataFrame,
                         ``genotype_df`` should have a column with this name and a numeric array type. If a column
                         expression is provided, the expression should return a numeric array type.
         dt : The numpy datatype to use in the linear regression test. Must be ``np.float32`` or ``np.float64``.
+        verbose_output: Whether or not to generate additional test statistics (n, sum_x, y_transpose_x)
+                        to the output DataFrame (default = False)
 
     Returns:
         A Spark DataFrame that contains
@@ -89,6 +91,9 @@ def linear_regression(genotype_df: DataFrame,
         - ``tvalue``: The T statistic
         - ``pvalue``: P value estimated from a two sided T-test
         - ``phenotype``: The phenotype name as determined by the column names of ``phenotype_df``
+        - ``n``: (verbose_output only) number of samples with non-null phenotype
+        - ``sum_x``: (verbose_output only) sum of genotype inputs, X
+        - ``y_transpose_x``: (verbose_output only) dot product of phenotype response and genotype input
     '''
 
     gwas_fx._check_spark_version(genotype_df.sql_ctx.sparkSession)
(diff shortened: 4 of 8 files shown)


bcajes:
style tests seem to be passing on my local environment, but not in circle ci.  It's not clear where the issue is from the logs.

python/glow/gwas/lin_reg.py
```diff
@@ -78,6 +79,8 @@ def linear_regression(genotype_df: DataFrame,
                         ``genotype_df`` should have a column with this name and a numeric array type. If a column
                         expression is provided, the expression should return a numeric array type.
         dt : The numpy datatype to use in the linear regression test. Must be ``np.float32`` or ``np.float64``.
+        verbose_output: Whether or not to generate additional test statistics (n, sum_x, y_transpose_x)
+                        to the output DataFrame (default = False)
```
karenfeng, line 83: Nit: remove `(default=False)`

python/glow/gwas/lin_reg.py
```diff
@@ -117,6 +113,7 @@ def linear_regression(genotype_df: DataFrame,
     Y = phenotype_df.to_numpy(dt, copy=True)
     Y_mask = (~np.isnan(Y)).astype(dt)
     Y = np.nan_to_num(Y, copy=False)
+    Y_raw_nan_filled = np.nan_to_num(Y, copy=True) if verbose_output else None
```
karenfeng, line 116: `Y` was already `nan`-filled; can we use `np.copy(Y)` [instead](https://numpy.org/doc/stable/reference/generated/numpy.copy.html)? Given this may also be `None`, it would also be clearer to name this variable something like `Y_for_verbose_output`.

python/glow/gwas/lin_reg.py
```diff
@@ -174,7 +167,8 @@ def _create_one_YState(Y: NDArray[(Any, Any), Float], phenotype_df: pd.DataFrame
 def _linear_regression_inner(genotype_pdf: pd.DataFrame, Y_state: YState,
                              Y_mask: NDArray[(Any, Any), Float], Y_scale: NDArray[(Any, ), Float],
                              Q: NDArray[(Any, Any), Float], dof: int,
-                             phenotype_names: pd.Series) -> pd.DataFrame:
+                             phenotype_names: pd.Series, Y_raw: NDArray[(Any, Any), Float] = None,
```
karenfeng, line 170: This is a private function, so we can change the APIs.

python/glow/gwas/lin_reg.py
```diff
@@ -201,6 +197,7 @@ def _linear_regression_inner(genotype_pdf: pd.DataFrame, Y_state: YState,
     pvalues = 2 * stats.distributions.t.sf(np.abs(T), dof)
 
     del genotype_pdf[_VALUES_COLUMN_NAME]
+
```
karenfeng, line 200: Can you revert the whitespace changes? It'll minimize unnecessary merge conflicts.

python/glow/gwas/lin_reg.py
```diff
@@ -209,5 +206,37 @@ def _linear_regression_inner(genotype_pdf: pd.DataFrame, Y_state: YState,
     out_df['tvalue'] = list(np.ravel(T))
     out_df['pvalue'] = list(np.ravel(pvalues))
     out_df['phenotype'] = phenotype_names.repeat(num_genotypes).tolist()
+    if verbose_output:
+        out_df["n"] = list(np.ravel(Y_mask.T @ np.ones(X.shape)))
```
karenfeng, line 210: Should we also calculate this in the earlier `if verbose_output` block to centralize these calculations?

karenfeng (commented):
Thanks for opening this PR @bcajes! The code looks clean overall. I have a high-level clarification question on the utility of the `verbose_output` argument - I understand that Hail provide `sum_x` and `y_transpose_x` (the latter of which is an intermediate calculations), but I'm not clear what they are used for.

python/glow/gwas/lin_reg.py
karenfeng: This is purely for consistency's sake; the documentation already contains the default.

bcajes:
Thanks for taking a look, Karen! It’s useful to have these stats side by side in some situations, as it gives the user (at a glance) a deeper understanding of the number of alt alleles, phenotypes, and interactions involved for each result .  In addition to this benefit, it’s a good sanity check as there are usually a series of transformations on phenotypes and genotypes that may be prone to error and that lead up to gwas.

python/glow/gwas/lin_reg.py
bcajes: I've removed the default values, but this function is still called in a unit test.

python/glow/gwas/lin_reg.py
bcajes: I think if I move some of the genotype_pdf optimization further upstream, I can do better centralize these calculations

karenfeng:
Thanks for addressing the comments @bcajes! In our case, these values are not calculated with data that are pre-processed via mean-centering and standardization (other than filling in `NaN`s) - my impression is that Hail expects the data to already be pre-processed. Does this align with your expectations as an end-user, given that we perform pre-processing otherwise? If so, it would be helpful to clarify this in the documentation.

It seems like you're having some issues with Python style tests - I would recommend running `yapfAll` in `sbt` to try to automatically fix the style. If that doesn't work, I can try to fix the errors in your PR manually.

bcajes:
Thanks, Karen. I think python/yapfAll did the trick, hopefully.  re: pre-processed data, yes, my expectation is that these values  correspond to the input data and not on standardization done within the function.

python/glow/gwas/lin_reg.py
```diff
@@ -88,6 +93,9 @@ def linear_regression(genotype_df: DataFrame,
         - ``tvalue``: The T statistic
         - ``pvalue``: P value estimated from a two sided T-test
         - ``phenotype``: The phenotype name as determined by the column names of ``phenotype_df``
+        - ``n``: (verbose_output only) number of samples with non-null phenotype
+        - ``sum_x``: (verbose_output only) sum of genotype inputs, X
+        - ``y_transpose_x``: (verbose_output only) dot product of phenotype response and genotype input
```
kianfar77, line 98: please mention the data structure of these new results. For example it seems y_transpose_x is a flattened array. Also need to explain precisely what happens to missing values in this output perhaps with an example.

python/glow/gwas/lin_reg.py
```diff
@@ -211,3 +210,32 @@ def _linear_regression_inner(genotype_pdf: pd.DataFrame, Y_state: YState,
...
+            StructField('y_transpose_x', sql_type)
+        ])
+
+    result_struct = gwas_fx._output_schema(genotype_df.schema.fields, result_fields)
+
+    def map_func(pdf_iterator):
+        for pdf in pdf_iterator:
+            yield gwas_fx._loco_dispatch(
+                pdf, Y_state, _linear_regression_inner, Y_mask, Y_scale, Q, dof,
+                phenotype_df.columns.to_series().astype('str'), Y_raw_nan_filled, verbose_output)
```
kianfar77, line 239: Is there a reason two different names are used in the code for this? `Y_raw_nan_filled` and `Y_for_verbose_output`?

kianfar77 (changes requested):
Thank your very much @bcajes for this PR. I left a few comments. Would be thankful if you address.

core/src/main/scala/io/projectglow/sql/expressions/VariantUtilExprs.scala
```diff
@@ -41,20 +41,25 @@ object VariantUtilExprs {
...
         .getStruct(i, genotypesSize)
         .getArray(genotypesIdx)
-      var sum = 0
-      var j = 0
-      while (j < calls.numElements() && sum >= 0) {
-        if (calls.getInt(j) >= 0) {
-          sum += calls.getInt(j)
-        } else {
-          sum = -1 // missing
+      if (calls == null) {
```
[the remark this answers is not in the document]
bcajes: Thanks for taking a look @kianfar77 .  I agree, it'll be cleaner to fix this issue further upstream.  I've added an ifnull clause to genotype call handling of hail entries.

python/glow/gwas/lin_reg.py
bcajes: There are no complex structures here, but hopefully an additional blurb and type info is sufficient.

python/glow/gwas/lin_reg.py
bcajes: No, we can use Y_for_verbose_output throughout

docs/README.md
```diff
@@ -1,4 +1,25 @@
...
+##### The Glow documentation is published with [Read the Docs](https://readthedocs.org/).
 
-To build the docs locally, activate the dev conda environment in `docs/source/environment.yml` and then
-run `make livehtml` from this directory.
+To build the docs locally,
+first create the conda environment:
+
+`conda env create -f source/environment.yml`
+
+activate the glow docs dev conda [environment](source/environment.yml):
```
henrydavidge, line 8: Thanks for this improvement!

docs/source/etl/variant-data.rst
```diff
@@ -72,7 +72,7 @@ You can control the behavior of the VCF reader with a few parameters. All parame
 
    Glow includes a VCF reader that uses `htsjdk <https://github.com/samtools/htsjdk>`_ for initial parsing as well as a reader that parses VCF lines to Spark rows directly.
 
-   As of release 0.7.0, the direct reader is enabled by default. To use the htsjdk based reader, set the Spark config ``io.projectglow.vcf.fastReaderEnabled`` to ``false``.
+   As of release 1.0.0, the direct reader is enabled by default. To use the htsjdk based reader, set the Spark config ``io.projectglow.vcf.fastReaderEnabled`` to ``false``.
```
henrydavidge, line 75: Think this should remain as is, perhaps with "As of" changed to "Since".

python/glow/gwas/lin_reg.py
```diff
@@ -191,6 +187,13 @@ def _linear_regression_inner(genotype_pdf: pd.DataFrame, Y_state: YState,
     '''
 
     X = np.column_stack(genotype_pdf[_VALUES_COLUMN_NAME].array)
+    del genotype_pdf[_VALUES_COLUMN_NAME]
+    num_genotypes = genotype_pdf.shape[0]
+    out_df = pd.concat([genotype_pdf] * Y_state.Y.shape[1])
+    if verbose_output:
+        out_df["n"] = list(np.ravel(Y_mask.T @ np.ones(X.shape)))
```
henrydavidge, line 194: nit: I think this would be both faster and easier to understand as `Y_mask.sum(axis=1).repeat(num_genotypes)` or something along those lines

python/glow/hail/tests/test_from_matrix_table.py
```diff
@@ -151,7 +151,7 @@ def test_plink(spark):
     matching_glow_df = glow_df.withColumn(
         'genotypes',
         fx.expr(
-            "transform(genotypes, gt -> named_struct('sampleId', gt.sampleId, 'calls', nullif(gt.calls, array(-1, -1)), 'phased', if(gt.calls = array(-1, -1), null, false)))"
+            "transform(genotypes, gt -> named_struct('sampleId', gt.sampleId, 'calls', ifnull(gt.calls, array(-1,-1)), 'phased', if(gt.calls = array(-1, -1), null, false)))"
```
henrydavidge, line 154: Why this change?

henrydavidge (approved):
Thanks for the contribution @bcajes! I had a few small comments. Looks great.

python/glow/gwas/lin_reg.py
bcajes: Thanks Henry, I've added this optimization

python/glow/hail/tests/test_from_matrix_table.py
bcajes: I wanted to make it explicit that the calls being compared should not be null.  May be it’ll be better as “’calls’,gt.calls” , but the comparison will fail due to differences in nullability. i.e. in hail/functions.py ifnull(e.GT.alleles, array(-1,-1)) is nullable=True.  I’m not sure if there’s an elegant way to change the nullability using the existing sql syntax.

bcajes:
There seems to be some issue with DCO and merge commits that I'm having trouble resolving

python/glow/hail/tests/test_from_matrix_table.py
karenfeng: @henrydavidge I believe this is related to the change in this PR to avoid NPEs where the `genotype.calls` array was `null` instead of `[-1, -1]`.

docs/README.md
williambrandler: hey this was me! @henrydavidge

kianfar77 (approved):
Thanks @bcajes for this PR.

PR #381, opened 2021-06-07
hail interoperability example

williambrandler:

add init script, job config, nb-test, notebooks, docs for hail interop

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

(Details)


williambrandler:
hey @karenfeng added in the hail example notebooks, initialization script and notebook tests are working. Please take a look. Thanks!

docs/dev/jobs-config-hail.json
```diff
@@ -0,0 +1,30 @@
+{
+  "new_cluster": {
+    "spark_version": "8.2.x-scala2.12",
```
karenfeng, line 3: This isn't guaranteed to work - Glow is built for Spark 3.0, and DBR 8.2 is on Spark 3.1.

docs/dev/jobs-config-hail.json
```diff
@@ -0,0 +1,30 @@
...
+    "node_type_id": "Standard_DS3_v2",
+    "num_workers": 2,
+    "spark_env_vars": {
+        "HAIL_VERSION": "0.2.65"
```
karenfeng, line 7: The version on which our tests are run is Hail 0.2.58. I believe that Hail upgraded directly from 2.4.5 to 3.1.1 in 0.2.65. We should not guarantee that this will work either.

docs/dev/jobs-config-hail.json
```diff
@@ -0,0 +1,30 @@
+{
```
karenfeng, line 1: This file is pretty similar to `docs/dev/jobs-config.json`. If we anticipate needing multiple different config files, it'd be cleaner if we had a minimal base JSON on which all configs are built, and diff JSONs that modify the base on an ad hoc basis.

docs/dev/jobs-config-hail.json
karenfeng: The safest thing would be a downgrade to DBR 6.4, which is supported until December; we also have Glow releases for Spark 2. We will need to cut a new Glow release to migrate to Spark 3.1.

docs/dev/run-nb-test.py
```diff
@@ -58,9 +61,14 @@ def main(cli_profile, workspace_tmp_dir, dbfs_init_script_dir, source_dir, nbs):
 
         print(f"Launching runs")
         for nb in nbs:
-            jobs_json['name'] = 'Glow notebook integration test - ' + nb
-            jobs_json['notebook_task'] = {'notebook_path': work_dir + '/' + nb}
-            run_submit = run_cli_cmd(cli_profile, 'runs', ['submit', '--json', json.dumps(jobs_json)])
+            if "hail" in nb:
```
karenfeng, line 64: It'd be cleaner to have:
```
jobs_json = jobs_json_base
if "hail" in nb:
  jobs_json = hail_jobs_json
jobs_json['notebook_task'] = {'notebook_path': work_dir + '/' + nb}
run_submit = run_cli_cmd(cli_profile, 'runs', ['submit', '--json', json.dumps(jobs_json)])
```

docs/dev/run-nb-test.py
```diff
@@ -17,6 +17,7 @@
 import uuid
 
 JOBS_JSON = 'docs/dev/jobs-config.json'
+JOBS_HAIL_JSON = 'docs/dev/jobs-config-hail.json'
```
karenfeng, line 20: `HAIL_JOBS_JSON = 'docs/dev/hail-jobs-config.json'`

docs/dev/run-nb-test.py
```diff
@@ -17,6 +17,7 @@
 import uuid
 
 JOBS_JSON = 'docs/dev/jobs-config.json'
```
karenfeng, line 19: Rename to `BASE_JOBS_JSON` and move the file to `docs/dev/base-jobs-config.json`, then add diff JSONs

docs/dev/run-nb-test.py
```diff
@@ -41,6 +42,8 @@ def main(cli_profile, workspace_tmp_dir, dbfs_init_script_dir, source_dir, nbs):
     work_dir = os.path.join(workspace_tmp_dir, identifier)
     with open(JOBS_JSON, 'r') as f:
         jobs_json = json.load(f)
+    with open(JOBS_HAIL_JSON, 'r') as f:
+        jobs_hail_json = json.load(f)
```
karenfeng, line 46: With the diff format, this would look like: `hail_jobs_json = base_jobs_json.update(json.load(f))`

karenfeng (changes requested):

docs/dev/jobs-config-hail.json
williambrandler: OK, let's wait until the new Glow release
Reluctant to have this run on Hail with Spark 2 and old versions of hail, 
since we have had people ask for hail and glow to work together with spark 3

docs/dev/jobs-config-hail.json
williambrandler: I think this makes sense if we were just changing variables such as the `spark_version`, but there are complicated changes to the json for Hail and VEP, such as new init scripts and docker containers. 

Also a user may just want to copy the json directly to use for their own cluster configuration. 

My preference is to add an arg to `run-nb-test.py` script that points to the json

WDYT?

docs/dev/run-nb-test.py
williambrandler: that does seem nicer, although I am leaning towards providing an arg to the python script instead, so we do not need `if "x" in nb:`

docs/dev/run-nb-test.py
williambrandler: how about

`python3 docs/dev/run-nb-test.py --cli-profile docs-ci --nbs etl/hail-interoperability --jobs-json-config hail`

`@click.option('--jobs-json-config', default='docs/dev/jobs-config.json', help='Configuration JSON file for Databricks jobs cluster.')`

`def main(cli_profile, workspace_tmp_dir, dbfs_init_script_dir, source_dir, nbs, jobs_json_config):
    JOBS_JSON = jobs_json_config`

docs/dev/jobs-config-hail.json
williambrandler: added this json and updated the script

williambrandler: tests passed first time! I'm getting the hang of this
@karenfeng let me know what you think

docs/dev/notebook-jobs-config-mapping.json
```diff
@@ -0,0 +1,19 @@
+{
+  "etl/gff": "docs/dev/jobs-config.json",
```
karenfeng, line 2: This adds a lot of boilerplate. Can we default to `docs/dev/jobs-config.json` if a job config is not present in the dictionary? Then the dictionary can contain only the custom cases.

karenfeng (changes requested):

Issue #387, opened 2021-06-29
Release on Spark 3.1


karenfeng:
Spark 3.1 was released on [March 2](https://spark.apache.org/releases/spark-release-3-1-1.html), we should rebase the glow-spark3 release to be on Spark 3.1 (currently on Spark 3.0).

dberma15:
This would be helpful as I'm having some troubles with glow on later versions.

williambrandler:
glow v1.1.0 has now been released, so closing this thread

PR #396, opened 2021-08-03
Upgrade Spark dependency from 3.0.0 to 3.1.2

karenfeng:

Works on issue #387 by upgrading Glow's Spark 3 build to use Spark 3.1.2.

This enables interoperability with Hail, whose Spark 3.x versions are built for Spark 3.1.

Follow up steps:

- Release Glow 1.1.0 with Spark 3.1.2 dependency
- Modify notebook tests to use Glow 1.1.0
- Merge in Hail notebooks

## How is this patch tested?
- [x] Unit tests
- [x] Integration tests
- [ ] Manual tests


python/glow/functions.py
```diff
@@ -490,10 +490,10 @@ def sample_call_summary_stats(genotypes: Union[Column, str], refAllele: Union[Co
 
     Examples:
         >>> sites = [
-        ... Row(refAllele='C', alternateAlleles=['G'], genotypes=[Row(sampleId='NA12878', calls=[0, 0])]),
-        ... Row(refAllele='A', alternateAlleles=['G'], genotypes=[Row(sampleId='NA12878', calls=[1, 1])]),
-        ... Row(refAllele='AG', alternateAlleles=['A'], genotypes=[Row(sampleId='NA12878', calls=[1, 0])])]
-        >>> df = spark.createDataFrame(sites, 'alternateAlleles: array<string>, genotypes: array<struct<sampleId: string, calls: array<int>>>, refAllele: string')
+        ... {'refAllele': 'C', 'alternateAlleles': ['G'], 'genotypes': [{'sampleId': 'NA12878', 'calls': [0, 0]}]},
```
karenfeng, line 493: PySpark stopped sorting Row fields in Spark 3.x, so this broke the tests. See https://spark.apache.org/docs/3.0.1/pyspark-migration-guide.html: we would have to set the environment variable PYSPARK_ROW_FIELD_SORTING_ENABLED to true for both executors and driver.

core/src/test/scala/io/projectglow/tertiary/AggregateByIndexSuite.scala
```diff
@@ -150,7 +150,7 @@ class AggregateByIndexSuite extends GlowBaseTest {
         """.stripMargin)
       .as[Seq[Int]]
       .head
-    // null if Spark 3.0, empty list if Spark 2.4
+    // null if Spark 3.x, empty list if Spark 2.4
```
kianfar77, line 153: Can we depend these directly on spark.version instead of OR?

python/glow/functions.py
kianfar77: We need to document this for developers.

kianfar77 (approved):
Thanks @karenfeng for this. Left a couple of comments. Otherwise great.

python/glow/functions.py
karenfeng: We should point to the migration guides in the release notes, but I'm not sure where else to document this. What do you think?

williambrandler: notebooks are not working with the new version of spark 3.1.2
Did you test the notebooks with this version of Spark?

When reading a VCF with Glow v1.1.0 and
`PYSPARK_ROW_FIELD_SORTING_ENABLED=true`

I get:

`java.lang.NoSuchMethodError: org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.transformAllExpressions(Lscala/PartialFunction;)Lorg/apache/spark/sql/catalyst/plans/logical/LogicalPlan;`

Any idea how to resolve this?

williambrandler: Also the notebook tests only test changed notebooks, but for a new version would want to change all of them?

williambrandler: this was caused by a regression in Spark, which will be resolved soon
https://databricks.atlassian.net/browse/SC-84055
will continue with the release once this is in

PR #402, opened 2021-09-01
Fix VEP parsing failures stemming from indels

### LINKED ISSUE #389: schema for VEP positional INFO fields incorrect for indels
The following fields for VEP are observed as IntegerType for SNPs, but for indels can be StringType

`
    "cDNA_position" -> IntegerType,
    "CDS_position" -> IntegerType,
    "Protein_position" -> IntegerType
`
This is because indels are specificed as a range of positions, for example, "48-49"
A consequence is that any VCF rows with indels will have their VEP INFO fields read in as "null"

https://github.com/projectglow/glow/blob/8b0bcd6b2f7320c3a5bd186bdcfa4707af303b47/core/src/main/scala/io/projectglow/vcf/AnnotationUtils.scala#L58

bboutkov:
Signed-off-by: Boris Boutkov <boris.boutkov@gmail.com>

Small collection of Int to String conversions for VEP parsing. Fixes #389 

## How is this patch tested?
- [ ] Unit tests
- [ ] Integration tests
- [ X] Manual tests


williambrandler (approved):
LGTM thanks @bboutkov 

This will resolve the "null" rows on annotation of indels, which represent position ranges as strings

williambrandler:
looks like the codecov failed because the test coverage dropped by `0.07%`
Will have to wait for either @karenfeng or @kianfar77 to come back from PTO and adjust this setting, as I do not have access. Therefore expect we will have to wait to merge this into glow v1.1.1

PR #416, opened 2021-09-30
Extending sample masking functionality in gwas linear regression

bcajes:

Added the following options and functionality to gwas linreg

intersect_samples: The current implementation of linear regression is optimized for speed,
                           but is not robust to missing phenotype values.  Without handling missingness
                           appropriately, pvalues may become inflated due to imputation.  When intersect_samples is
                           enabled, samples that do no exist in the phenotype dataframe will be dropped from
                           genotypes, offsets, and covariates prior to regression analysis.  Note that if phenotypes in
                           phenotypes_df contain missing values, these samples will not be automatically dropped.
                           The user is responsible for determining their desired levels of missingness and imputation.
                           Drop any rows with missing values from phenotype_df prior to linear_regression to prevent
                           any imputation.  If covariates are provided, covariate and phenotype samples will
                           automatically be intersected.

genotype_sample_ids: Sample ids from genotype_df.
                             i.e. from applying glow.wgr.functions.get_sample_ids(genotype_df) or
                             if include_sample_ids=False was used during the generation genotype_df, then using an
                             externally managed list of sample_ids that correspond to the array of genotype calls.

The alternative to using these options is to perform these pre-processing steps prior to calling linear_regression.  Outside of a lot of additional boilerplate code that the alternative approach will incur, transforming the genotype_df in spark (as opposed to masking operations implemented in this change) can be very expensive.

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

Issues with inflation can be replicated by simply adding missing values to the phenotype_df and disabling intersect_samples.
Tested these changes on real world and simulated data and compared results to hail and statsmodel implementations to ensure pvalues and betas were concordant.


python/glow/gwas/lin_reg.py
```diff
@@ -23,11 +23,13 @@ def linear_regression(genotype_df: DataFrame,
...
                       dt: type = np.float64,
-                      verbose_output: bool = False) -> DataFrame:
+                      verbose_output: bool = False,
+                      intersect_samples: bool = False,
+                      genotype_sample_ids: Optional[List[str]] = None) -> DataFrame:
     '''
     Uses linear regression to test for association between genotypes and one or more phenotypes.
     The implementation is a distributed version of the method used in regenie: 
-    https://www.nature.com/articles/s41588-021-00870-7
+    https://www.biorxiv.org/content/10.1101/2020.06.19.162354v2
```
williambrandler, line 32: Please keep this link pointing to the published version of the article in nature genetics

python/glow/gwas/lin_reg.py
```diff
@@ -23,11 +23,13 @@ def linear_regression(genotype_df: DataFrame,
                       add_intercept: bool = True,
                       values_column: Union[str, Column] = 'values',
                       dt: type = np.float64,
-                      verbose_output: bool = False) -> DataFrame:
+                      verbose_output: bool = False,
+                      intersect_samples: bool = False,
```
williambrandler, line 27: Do you think we should set `intersect_samples: bool = True` by default?

python/glow/gwas/lin_reg.py
```diff
@@ -83,6 +85,20 @@ def linear_regression(genotype_df: DataFrame,
...
+                           but is not robust to high levels missing phenotype values.  Without handling missingness
+                           appropriately, pvalues may become inflated due to imputation.  When intersect_samples is
+                           enabled, samples that do no exist in the phenotype dataframe will be dropped from
+                           genotypes, offsets, and covariates prior to regression analysis.  Note that if phenotypes in
+                           phenotypes_df contain missing values, these samples will not be automatically dropped.
+                           The user is responsible for determining their desired levels of missingness and imputation.
+                           Drop any rows with missing values from phenotype_df prior to linear_regression to prevent
+                           any imputation.  If covariates are provided, covariate and phenotype samples will
+                           automatically be intersected.
+        genotype_sample_ids: Sample ids from genotype_df.
```
williambrandler, line 98: nice

williambrandler (commented):
LGTM, let's get one final sign off from @henrydavidge before merging in. 

Then I will get started on adding in the integration test notebooks for linear regression

python/glow/gwas/lin_reg.py
```diff
@@ -186,14 +220,18 @@ def _linear_regression_inner(genotype_pdf: pd.DataFrame, Y_state: YState,
     So, if a matrix's indices are `sg` (like the X matrix), it has one row per sample and one column per genotype.
     '''
 
-    X = np.column_stack(genotype_pdf[_VALUES_COLUMN_NAME].array)
+    genotype_values = genotype_pdf[_VALUES_COLUMN_NAME].array
+    if gt_indices_to_drop is not None and gt_indices_to_drop.size:
+        genotype_values = list(map(lambda x: np.delete(x, gt_indices_to_drop), genotype_values))
+    X = np.column_stack(genotype_values)
```
henrydavidge, line 226: You can mask out the samples after this line like `np.delete(X, gt_indices_to_drop, axis=0)`. It's best to avoid pure Python loops whenever possible.

python/glow/gwas/log_reg.py
```diff
@@ -290,7 +303,10 @@ def _logistic_regression_inner(genotype_pdf: pd.DataFrame, log_reg_state: LogReg
     p: phenotype
     c, d: covariate
     '''
-    X = np.column_stack(genotype_pdf[_VALUES_COLUMN_NAME].array)
+    genotype_values = genotype_pdf[_VALUES_COLUMN_NAME].array
+    if gt_indices_to_drop is not None and gt_indices_to_drop.size:
+        genotype_values = list(map(lambda x: np.delete(x, gt_indices_to_drop), genotype_values))
```
henrydavidge, line 308: Same comment here.

henrydavidge (approved):
Assuming that you have tested at scale on your data and everything looks good, LGTM! Thanks a lot @bcajes.

python/glow/gwas/lin_reg.py
williambrandler: actually let's keep it false, and I will make a quickstart guide that demos this functionality, showing confounded results vs non-confounded (with `intersect_samples` set to True)

PR #418, opened 2021-10-06
Iss375 observability

mah-databricks:
Basic Observability for CLI tooling called from Glow


core/src/main/scala/io/projectglow/transformers/pipe/Piper.scala
```diff
@@ -208,6 +216,12 @@ class PipeIterator(
     } else {
       val exitStatus = processHelper.waitForProcess()
       if (exitStatus != 0) {
+        quarantineInfo.foreach{quarantineInfo =>
+          val th = processHelper.childThreadException.getOrElse(
+            new Throwable("unknown"))
```
dmoore247, line 221: Could we have something more descriptive here?

core/src/main/scala/io/projectglow/transformers/pipe/Piper.scala
```diff
@@ -60,10 +60,13 @@ private[projectglow] object Piper extends GlowLogging {
       outputformatter: OutputFormatter,
       cmd: Seq[String],
       env: Map[String, String],
-      df: DataFrame): DataFrame = {
-
+      df: DataFrame,
+      quarantineLocation: Option[String] = None): DataFrame = {
```
dmoore247, line 64: Is this a path or a table? quarantineLocation name implies path, code does saveAsTable() which would be more accessible. Perhaps find a way to document / indicate better this is a table rather than a path.

williambrandler:
hey, @mah-databricks thanks for adding the tests

Looks like the circleci checks are failing due to formatting, would you be able to resolve?

`[warn] /home/circleci/glow/core/src/main/scala/io/projectglow/transformers/pipe/Piper.scala isn't formatted properly!`

williambrandler (approved):
LGTM

PR #469, opened 2021-12-23
Iss463 quarantine full ds

mah-databricks:
Rework Piper to quarantine only the partitions which fail. This required a small rewrite to the Iterator code which calls the external CLI tooling. Also needed to backfill exceptions to make sure the application gives the appearance of still functioning the same way.

## How is this patch tested?
- [x ] Unit tests
- [ ] Integration tests
- [ x] Manual tests

(Details)


williambrandler:
thanks @mah-databricks, let me merge in, rebuild the project and then perform an integration test on this new quarantine functionality

PR #513, opened 2022-06-30
Test recalibration + DBR 9.1 Docker images

a0x8o:
1. Temporarily stop Hail tests (until these are upgraded for where breaking Hail changes are presently failing)
2. Fix broken links in Getting Started document source for Databricks Repos Sync/Git Clone 
3. Provide Docker images for DBR 9.1 

## How is this patch tested?
- [X] Unit tests
- [X] Integration tests
- [X] Manual tests

I am aiming to address three things here: 
1. Hail-IS made some changes that are breaking the Hail tests in projectglow. I'm switching off these tests until someone can fix the Hail tests.
2. Getting Started had another broken link which I've fixed. I also have found that these tests should have a higher timeout value as frequently the tests fail due to short load timeouts. 
3. I believe we need to provide the DBR9.1 Docker images in addition to DBR10.4 images in master. I've reintroduced these to master.

### DIFF
--- a/docker/databricks/build.sh
+++ b/docker/databricks/build.sh
@@ -3,6 +3,32 @@
 #
 # Usage: ./build.sh
 
+DATABRICKS_RUNTIME_VERSION="9.1"
+GLOW_VERSION="1.1.2"
+HAIL_VERSION="0.2.85"
+
+# build 9.1 LTS / Spark 3.1.2 images 
+
+# Add commands to build images below
+pushd dbr/dbr$DATABRICKS_RUNTIME_VERSION/
+docker build -t "${DOCKER_HUB}/minimal:${DATABRICKS_RUNTIME_VERSION}" minimal/
+docker build -t "${DOCKER_HUB}/python:${DATABRICKS_RUNTIME_VERSION}" python/
+docker build -t "${DOCKER_HUB}/dbfsfuse:${DATABRICKS_RUNTIME_VERSION}" dbfsfuse/
+docker build -t "${DOCKER_HUB}/standard:${DATABRICKS_RUNTIME_VERSION}" standard/
+docker build -t "${DOCKER_HUB}/with-r:${DATABRICKS_RUNTIME_VERSION}" r/
+docker build -t "${DOCKER_HUB}/genomics:${DATABRICKS_RUNTIME_VERSION}" genomics/
+docker build -t "${DOCKER_HUB}/databricks-hail:${HAIL_VERSION}" genomics-with-hail/
+docker build -t "${DOCKER_HUB}/databricks-glow-minus-ganglia:${GLOW_VERSION}" genomics-with-glow/
+docker build -t "${DOCKER_HUB}/databricks-glow:${GLOW_VERSION}" ganglia/
+docker build -t "${DOCKER_HUB}/databricks-glow-minus-ganglia:${DATABRICKS_RUNTIME_VERSION}" genomics-with-glow/
+docker build -t "${DOCKER_HUB}/databricks-glow:${DATABRICKS_RUNTIME_VERSION}" ganglia/
+popd
+
+docker push "${DOCKER_HUB}/databricks-hail:${HAIL_VERSION}"
+docker push "${DOCKER_HUB}/databricks-glow:${GLOW_VERSION}"
+
+# build 10.4 LTS / Spark 3.2.1 images
+
 DOCKER_HUB="projectglow"
 DATABRICKS_RUNTIME_VERSION="10.4"
 GLOW_VERSION="1.2.1"
@@ -26,3 +52,7 @@ popd
 docker push "${DOCKER_HUB}/databricks-hail:${HAIL_VERSION}"
 docker push "${DOCKER_HUB}/databricks-glow:${GLOW_VERSION}"
 docker push "${DOCKER_HUB}/databricks-glow:${DATABRICKS_RUNTIME_VERSION}"
+
+
+
+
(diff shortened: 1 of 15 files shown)


a0x8o:
@williambrandler @karenfeng I am trying to address three things here: 
1. Hail-IS made some changes that are breaking the Hail tests in Projectglow. I'm switching off these tests until someone (I or another engineer) can fix the tests.
2. Getting Started had another broken link which I've fixed. I also have found that these tests should have a higher timeout value as frequently the tests fail due to short load timeouts. 
3. I believe we need to provide the DBR9.1 Docker images in addition to DBR10.4 images in master.

williambrandler:
ah ok, thanks @a0x8o 

I removed 9.1 when bumping to 10.4, but it seems reasonable to maintain both versions until 9.1 is obsolete

Fine to remove hail tests to get the ci passing again, can add back in once https://github.com/hail-is/hail/issues/11707 is resolved. The Hail team are waiting for EMR and dataproc to update to spark 3.2.0, after which I expect the tests to pass (once the Hail version is updated)

I see [EMR](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/Spark-release-history.html) is on 3.2.0 with emr-6.6.0, however dataproc is still on spark 3.1.3 [dataproc](https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-release-2.0)

docker/databricks/build.sh
```diff
@@ -3,6 +3,32 @@
...
 # Usage: ./build.sh
 
+DATABRICKS_RUNTIME_VERSION="9.1"
+GLOW_VERSION="1.1.2"
+HAIL_VERSION="0.2.85"
+
+# build 9.1 LTS / Spark 3.1.2 images 
+
+# Add commands to build images below
+pushd dbr/dbr$DATABRICKS_RUNTIME_VERSION/
```
williambrandler, line 13: could add a loop here instead of having the commands twice,

something like this: https://stackoverflow.com/questions/35543088/iterating-over-list-of-arrays-in-bash

docker/databricks/dbr/dbr9.1/genomics-with-glow/Dockerfile
```diff
@@ -0,0 +1,57 @@
+# ===== For the runtime environment for this image we need the databricks azure setup ============== 
+
+FROM projectglow/genomics:9.1 AS builder
+
+ENV GLOW_VERSION=1.1.1
```
williambrandler, line 5: 1.2.1

docker/databricks/dbr/dbr9.1/genomics-with-hail/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# ===== For the runtime environment for this image we need the databricks azure setup ==============
+
+FROM projectglow/genomics:9.1 AS builder
+
+# ===== Set up Hail ================================================================================
+
+# earliest Hail version supported by Spark3 is 0.2.67
+ENV HAIL_VERSION=0.2.78
```
williambrandler, line 8: match to HAIL_VERSION in build script

docker/databricks/dbr/dbr9.1/genomics/Dockerfile
```diff
@@ -0,0 +1,115 @@
+FROM projectglow/with-r:9.1 AS r
```
williambrandler, line 1: some of these files are identical between dbr9.1 and dbr10.4, 

is there a clever way to reference unchanged files without duplication?

williambrandler (commented):
it would be more elegant if there are ways to avoid duplication of files and of commands in the bash script

docker/databricks/dbr/dbr9.1/genomics-with-glow/Dockerfile
a0x8o: Shouldn't this be 1.1.2? 1.2.1 was the version upgraded for Spark 3.2.1 (vs. 1.1.2 supporting Spark 3.1.2)

docker/databricks/dbr/dbr9.1/genomics-with-hail/Dockerfile
a0x8o: Shouldn't this be 0.2.78 as this was tested on DBR 9.1 LTS?

a0x8o:
@williambrandler any idea why we are seeing the following:
https://app.circleci.com/pipelines/github/projectglow/glow/3131/workflows/a1e5ccf6-ada6-480b-91d8-9983139c4cd9/jobs/9682

docker/databricks/dbr/dbr9.1/genomics/Dockerfile
williambrandler: ok sounds good

docker/databricks/dbr/dbr9.1/genomics-with-hail/Dockerfile
williambrandler: [build.sh](https://github.com/projectglow/glow/pull/513/files/9a095b32e7efc747ccf3e81345c37be0edcc44da#diff-5e88d12e5231077937292f2bf9dd8b156fd542f46ecb353fe52ced01931ff1f9) has the hail version tagged as `0.2.85`, please change that script to `0.2.78` to match this Dockerfile

williambrandler:
> @williambrandler any idea why we are seeing the following: https://app.circleci.com/pipelines/github/projectglow/glow/3131/workflows/a1e5ccf6-ada6-480b-91d8-9983139c4cd9/jobs/9682

either a transient issue with circleci or permissions related @a0x8o

Try this link to join the project?
https://app.circleci.com/pipelines/github/projectglow/glow?invite=true

PR #546, opened 2024-01-17
Update to Spark 3.4/3.5

henrydavidge:
- Update supported Spark versions to 3.4 and 3.5
- Remove Hail integration
- Remove features that frequently cause incompatibilities between versions (`aggregate_by_index`, CSV pipe transformer). Workarounds will be documented in a follow-up PR.
- Update Scala and Python dependencies to match recent DBR releases
- Fix unit tests that are failing in master
- Clean up test logs

## How is this patch tested?
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

(Details)


henrydavidge:
Merging since all tests pass (except for the known-broken notebook tests)

henrydavidge:
@a-li There will definitely be a maven release, but I'm planning on adding a few features so it might be a couple weeks. If you want to try it out before then, you follow these instructions: https://github.com/projectglow/glow?tab=readme-ov-file#testing-code-on-a-databricks-cluster. The artifacts built this way can be used anywhere, not just in Databricks.

PR #660, opened 2024-04-19
Setting upper threshold on pandas <2.0.0

zyd14:
Pinning pandas to < 2.0.0 in setup.py

## How is this patch tested?
- [X] Unit tests

(Describe any other testing)

To run Spark 4.0 tests, add `[SPARK4]` to the pull request title.


henrydavidge:
cc @kermany I think it would be better to replace the uses of `append` with `pd.concat` so that the code works with both major pandas versions. As a workaround you should also be able to explicitly install a pandas version <2.0.0.

kermany:
@henrydavidge agreed, we should change it to append (I will create a PR) in the meanwhile approve this.

henrydavidge:
Yep, makes sense to me. Thanks @zyd14 @kermany

